PR #85, opened 2026-04-06
Auto-fit Column Width

gepardjaro:
What this adds

  - Double-click any column header → auto-fits that column to the widest content in the loaded log
  - ↔ button in the severity column header → auto-fits all visible columns at once
  - Both changes are permanent and saved to preferences (same as a manual drag-resize)

  ---
  Files changed

  src/lib/log-accessibility.ts

  Added getCanvasFont(size, bold?, sourceElement?) — returns a canvas-compatible font string. Prefers reading fontFamily from a real rendered
  DOM element, falls back to resolving the --cmtrace-font-family-ui CSS custom property. This is needed because canvas.measureText() cannot
  handle var(--...) syntax.

  src/lib/column-config.ts

  Added three exports:

  - measureTextWidth(text, font) — measures pixel width of a string using a singleton canvas context (avoids repeated DOM creation).
  - calcAutoFitWidth(col, entries, contentFont, headerFont) — computes the optimal column width:
    - severity → returns its fixed defaultWidth (renders a dot, not text)
    - dateTime → uses a representative timestamp string ("2024-01-01 00:00:00.000") since the view layer formats it separately
    - All other columns → two-pass approach:
        - Pass 1 (O(n), no canvas): scans every entry by String.length to find the maximum string length across the entire dataset — no
  MAX_SAMPLE cap, so entries at line 50,000+ are always included
      - Pass 2 (O(k), canvas): runs measureText() only on strings within 90% of the maximum length, keeping canvas calls minimal
    - If no data is found for a column, returns defaultWidth (never shrinks an empty column below its designed size)
    - Result is Math.ceil'd to prevent sub-pixel under-sizing

  src/components/log-view/LogListView.tsx

  - Added imports: ArrowBidirectionalLeftRightRegular from @fluentui/react-icons, calcAutoFitWidth from column-config, getCanvasFont from
  log-accessibility
  - handleHeaderDoubleClick(colId) — resolves font from a rendered .log-row DOM element via parentRef (guarantees the same font the browser is
  actually using), calls calcAutoFitWidth, then setColumnWidth
  - handleFitAllColumns() — same font resolution, iterates all visibleColumns and sets each
  - HeaderCellProps / HeaderCell — added onDoubleClick: (colId) => void and optional onFitAll?: () => void props
  - HeaderCell renders the ArrowBidirectionalLeftRightRegular icon button when onFitAll is provided (passed only to the severity column, which
  has no label and is always first), otherwise renders the normal column label
  - onDoubleClick is attached to the header cell div; e.preventDefault() prevents any drag/selection conflict
  - Header strip structure reverted to original (no extra wrapper div, no absolute overlay) — the button lives naturally inside the severity
  cell

### DIFF
--- a/src/lib/log-accessibility.ts
+++ b/src/lib/log-accessibility.ts
@@ -45,6 +45,29 @@ export function getLogListMetrics(fontSize: number): LogListMetrics {
   };
 }
 
+/**
+ * Returns a canvas-compatible font string.
+ * Prefers reading the resolved font-family from a real DOM element (most accurate),
+ * falling back to resolving the CSS custom property, then a hardcoded fallback.
+ * canvas.measureText() cannot handle var(--...) syntax.
+ */
+export function getCanvasFont(
+  size: number,
+  bold = false,
+  sourceElement?: Element | null
+): string {
+  let family: string;
+  if (sourceElement) {
+    family = getComputedStyle(sourceElement).fontFamily;
+  } else {
+    family =
+      getComputedStyle(document.documentElement)
+        .getPropertyValue("--cmtrace-font-family-ui")
+        .trim() || "'Segoe UI', Tahoma, sans-serif";
+  }
+  return `${bold ? "bold " : ""}${size}px ${family}`;
+}
+
 export function getLogDetailsLineHeight(fontSize: number): number {
   const clampedFontSize = clampLogDetailsFontSize(fontSize);
   return Math.max(20, Math.round(clampedFontSize * 1.6));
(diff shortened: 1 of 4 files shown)


src/lib/log-accessibility.ts
adamgell: this is also doing something with fonts?

adamgell (commented):
Thanks for the contribution! The auto-fit feature itself looks good, but could you split this PR so it only includes the files related to the column auto-fit feature? The CLAUDE.md changes (test commands, lite build, cargo deny, event_log module, etc.) are unrelated and should go in a separate PR.

PR #180, opened 2026-04-29
Fix review findings

adamgell:
## Summary
- Fix tail partial-record buffering so incomplete trailing records are not duplicated on the next read.
- Make frontend filter memoization source-aware so filters refresh across file/tab changes.
- Reset tab state before session restore to avoid mixing old and restored tabs.
- Validate and parse date/time filter values instead of silently matching all rows.
- Reset folder parse progress counters at the start of each new load.

## Follow-up review fixes
- `parse_filter_timestamp_millis` now also accepts whole-second datetimes (`2026-04-01 12:30:00`, `2026-04-01T12:30:00`, `04/01/2026 12:30:00`); chrono's `%.f` requires the leading dot, so the original list silently rejected these (Copilot review comment).
- Naive datetime filter inputs are interpreted as **local time** (`chrono::Local::from_local_datetime`) instead of UTC, matching the frontend's local-time display.
- `apply_filter` compiles clauses once before the per-entry loop (pre-parses timestamp targets and lower-cases string needles), avoiding O(entries) repeat work on large logs.
- `match_timestamp` is a single exhaustive match — the previously-dead `Contains | NotContains` arm after the early return is gone; each comparison op parses lazily.
- `useParseProgressListener` subscribes to a derived `isFolderLoading` boolean instead of the raw `folderLoadProgress` number, eliminating per-progress-tick re-renders during folder parses.
- New `useUiStore.clearTabs()` action replaces a raw `setState` shallow-merge in `restoreSession`, also clears active log + filter state to mirror `closeTab`'s last-tab path, and the per-file restore now falls back to the aggregate loader only when every individual file failed.
- `restoreSession` resolves `activeTabIndex` and per-tab scroll restoration against the live `useUiStore.openTabs` list, so partial load failures don't misalign the index.
- Extracted `lib/tab-snapshot-cache.ts` (`TabEntrySnapshot`, `SourceOpenMode`, `get/set/clear/clearAll` helpers); `ui-store` imports directly from there instead of going through `log-store`, breaking that piece of the store-to-store cycle.

## Marker-key fix (out-of-scope but caught during review)
- Marker context-menu had been writing under `entry.filePath` since #126, but `LogListView` still keyed reads off a tab-level `activeFilePath`. In aggregate-folder mode markers were saved correctly but never displayed for rows whose source file wasn't `entries[0].filePath`, so "Mark as Bug" / "Mark as Confirmed" appeared to do nothing.
- LogRow callbacks now take `(filePath, lineId)`; LogListView resolves markers per row via `markersByFile.get(entry.filePath)?.get(entry.id)`, loads markers for every unique `filePath` in `displayEntries`, and tracks a per-file dirty set for the debounced save.

## Column auto-fit (out-of-scope but caught during review)
- The `message` column was excluded from auto-fit entirely, so on logs where Log Text is the dominant column (macOS install.log, generic timestamped logs) the Fit-All button and the divider double-click appeared to be no-ops. Message now uses the same max-content-width logic as every other column, capped at the existing `MAX_AUTOFIT_WIDTH` (1200px).
- Double-click now fits the column when fired anywhere on the header, not only the 10px resize handle on the right edge — the affordance was undiscoverable. Verified end-to-end against real `CcmExec.log` data via Playwright (1200px cap reached on long messages).

## Verification
- npx tsc --noEmit
- npm test (104 tests pass)
- cargo check
- cargo test (incl. new `timestamp_filter_accepts_whole_second_datetimes`, `timestamp_filter_naive_input_is_local_time`, `compile_clauses_*`)
- cargo clippy -- -D warnings

Notes: no ESLint config was present in the repository.

### DIFF
--- a/src-tauri/src/commands/filter.rs
+++ b/src-tauri/src/commands/filter.rs
@@ -36,6 +36,8 @@ pub struct FilterClause {
 
 /// Apply filter clauses to a list of entries.
 /// Returns the IDs of entries that match ALL clauses (AND logic).
+/// Returns `Err(AppError::InvalidInput)` if a timestamp clause has an
+/// unparseable value so the caller can surface it to the user.
 #[tauri::command]
 pub fn apply_filter(
     entries: Vec<LogEntry>,
@@ -79,7 +81,7 @@ fn matches_clause(entry: &LogEntry, clause: &FilterClause) -> Result<bool, crate
         }
         FilterField::Timestamp => {
             let ts = entry.timestamp.unwrap_or(0);
-            return match_timestamp(ts, &clause.op, &clause.value);
+            match_timestamp(ts, &clause.op, &clause.value)?
         }
         FilterField::Severity => {
             let sev_str = match &entry.severity {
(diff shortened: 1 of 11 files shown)


adamgell:
### Code review

Found 1 issue:

1. `parse_filter_timestamp_millis` rejects common ISO-8601 timestamps without fractional seconds. All four `NaiveDateTime` formats listed include `%.f`, which in chrono is **not** optional — the leading dot is required. Inputs like `2026-04-01 12:30:00` or `2026-04-01T12:30:00` (no fraction, no timezone) fall through every branch (RFC 3339 also requires a timezone designator) and end up returning `AppError::InvalidInput`, surfacing a hard error to the UI for what was previously a silently-passing filter. This matches the Copilot inline comment at https://github.com/adamgell/cmtraceopen/pull/180#discussion_r3164249739 — recommend adding `"%Y-%m-%dT%H:%M:%S"` and `"%Y-%m-%d %H:%M:%S"` (and the equivalent for `%m/%d/%Y %H:%M:%S`) before the `%.f` variants, or making the fraction optional.

https://github.com/adamgell/cmtraceopen/blob/a459c63d7df3f0add839b69b26e987fea7b3565e/src-tauri/src/commands/filter.rs#L140-L151

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #236, opened 2026-07-09
ci: alert on nightly signed build failures

adamgell:
## What

Two small jobs added to the nightly signed build workflow:

- **`notify-failure`** — on a failed **scheduled** run, opens a tracking issue (label `nightly-build-failure`), or comments on the existing open one, with the run URL + commit. Gated to `schedule` so manual dispatches don't spam issues.
- **`notify-recovery`** — on any fully green run, closes the open tracking issue with a recovery note. A manual green run also clears the alert, so the issue self-maintains.

Both use the built-in `GITHUB_TOKEN` (`issues: write`) — no external secrets, no checkout.

## Why

The signed nightly failed silently for ~34 nights (Azure Artifact Signing broke after a subscription tenant move) before it was noticed. This makes a month-long outage impossible to miss.

## Validation

- `actionlint` clean
- YAML parses; all 6 jobs load

### DIFF
--- a/.github/workflows/cmtrace-nightly-signed.yml
+++ b/.github/workflows/cmtrace-nightly-signed.yml
@@ -745,8 +745,8 @@ jobs:
             --description "Automated: signed nightly build failure" \
             || gh label edit "$label" --color b60205 --description "Automated: signed nightly build failure"
 
-          existing="$(gh issue list --label "$label" --state open --json number \
-            --jq '.[0].number // empty' 2>/dev/null || true)"
+          existing="$(gh issue list --label "$label" --state open --limit 1 --json number \
+            --jq '.[0].number // empty')" || exit $?
 
           if [ -n "$existing" ]; then
             gh issue comment "$existing" \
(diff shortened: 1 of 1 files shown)


adamgell (commented):
Reviewed head `3d2f996a6a8` against `main`.

No findings in the PR diff. The workflow logic matches the stated failure/recovery requirements, all three prior review threads are resolved/outdated, and local validation passed: `git diff --check`, YAML parse (6 jobs), `actionlint`, `bash -n`, ShellCheck, and CodeRabbit (0 findings).

Do not merge yet: required `Check & Test (Rust)` is failing in `cargo deny` on newly published `RUSTSEC-2026-0194` and `RUSTSEC-2026-0195` for transitive `quick-xml 0.39.4`; required build checks were consequently skipped. That failure is external to this YAML-only diff, but it should not be bypassed. Re-run to full green after the advisory/dependency policy is resolved, then obtain the required code-owner approval.

PR #246, opened 2026-07-11
fix(ci): make cargo-audit pass by ignoring transitive-dep vulnerabilities

adamgell:
The required `Check & Test (Rust)` gate has been failing repo-wide at its final step, `Security vulnerability audit` (`cargo audit`) — which also blocks the platform `build` jobs (they `need` this job). This unblocks it.

## Why deny.toml alone wasn't enough
`cargo audit` and `cargo deny check` are two separate tools. `cargo deny check` reads `src-tauri/deny.toml` (and passes). **`cargo audit` does not read `deny.toml`** — it reads `.cargo/audit.toml`, which didn't exist. So the advisory ignores only ever applied to `cargo deny`, and `cargo audit` failed on every run.

`cargo audit` reported **5 vulnerability findings across 3 advisories**, all in transitive dependencies that can't be upgraded without a Tauri stack bump:

| Advisory | Crate | Path |
|----------|-------|------|
| RUSTSEC-2026-0194 | quick-xml | plist/Tauri, wayland-scanner |
| RUSTSEC-2026-0195 | quick-xml | plist/Tauri, wayland-scanner |
| RUSTSEC-2026-0204 | crossbeam-epoch | transitive |

(The other 22 findings are unmaintained-crate *warnings* — `cargo audit` never failed on those.)

## Change
- Add repo-root **`.cargo/audit.toml`** with the 3 advisory ignores (this is the file `cargo audit` reads).
- Add the missing **RUSTSEC-2026-0204** to `src-tauri/deny.toml` so the two ignore lists stay in sync (the quick-xml pair was already there).

## Verification
Both exit 0 locally:
- `cargo audit` → `warning: 22 allowed warnings found` (no errors)
- `cargo deny check` → `advisories ok, bans ok, licenses ok, sources ok`

## Note on overlap with `fix/workspace-lockfile-sync`
That branch already introduces `.cargo/audit.toml`, but with only the two quick-xml IDs — it's now **missing RUSTSEC-2026-0204** (published more recently), so it would no longer fully pass on its own. This PR is the minimal standalone unblock; if it merges first, the lockfile-sync branch should take this file's 3-ID version (or drop its own copy). Merge order is a trivial add/add on one file.

Once this lands on `main`, updating the other open PRs (#241–#245) with `main` turns their audit step green and lets the platform builds run.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,14 @@
+# cargo-audit configuration.
+#
+# cargo-audit does NOT read src-tauri/deny.toml, so vulnerability advisories
+# that cannot be resolved by a version bump must be listed here as well.
+# Keep this list in sync with the [advisories] ignore list in src-tauri/deny.toml.
+#
+# All entries are transitive dependencies we do not control and cannot upgrade
+# without an upstream Tauri stack bump.
+[advisories]
+ignore = [
+    "RUSTSEC-2026-0194", # quick-xml <0.41: quadratic runtime on duplicate attrs (transitive: plist/Tauri, wayland-scanner)
+    "RUSTSEC-2026-0195", # quick-xml <0.41: unbounded namespace-declaration alloc (transitive: plist/Tauri, wayland-scanner)
+    "RUSTSEC-2026-0204", # crossbeam-epoch: invalid pointer deref in fmt::Pointer impl (transitive)
+]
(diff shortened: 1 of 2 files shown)


adamgell:
Superseded by #240, which already landed `.cargo/audit.toml` and synced Cargo.lock — `cargo audit` now passes on main (the crossbeam-epoch advisory dropped out with the lockfile resync). Closing as redundant.

PR #247, opened 2026-07-11
chore: upgrade to TypeScript 7 (GA) + bring changelog current since 1.3.2

adamgell:
## What

Upgrade the frontend's TypeScript from **6.0.3 → 7.0.2** — the native (Go) compiler, which reached GA on 2026-07-09. `typescript@7` ships `tsc` as the native binary.

## Does anything break? No.

With **zero source changes**:

| Check | Result |
|-------|--------|
| `tsc --noEmit` | ✅ **0 errors** |
| `tsc && vite build` (`npm run frontend:build`) | ✅ builds |
| `vitest` (`npm run test`) | ✅ **123/123** across 14 files |

## Does anything get faster? Yes — type-checking ~5.5×.

`tsc --noEmit`, warm, macOS arm64 (2 runs each):

| | TypeScript 6.0.3 | TypeScript 7.0.2 |
|---|---|---|
| `tsc --noEmit` | ~3.79 s | **~0.69 s** |

This speeds up both CI's **TypeScript Check** job and the local `frontend:build` script. vite and vitest transpile via esbuild and are unaffected — `tsc` is used only as the type-check gate.

## Notes

- **Lockfile growth** is expected: the native compiler distributes as per-platform binary packages (`@typescript/typescript-<os>-<arch>`) via `optionalDependencies`, resolved per-platform on install (same model as esbuild/swc). All CI targets — `linux-x64`, `win32-x64`, `darwin-arm64/x64` — are present in the lock, so `npm ci` resolves correctly everywhere.
- No `tsconfig.json` changes needed; the native compiler consumes the existing config.
- Node requirement is `>=16.20`; CI uses Node 20. ✓

## How this was tested

Isolated worktree off `main`; measured TS 6 baseline, upgraded, re-measured, then ran the full build + test suite. CI (Linux/Windows/macOS) will confirm cross-platform.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- **TypeScript 7 (#247)**: Upgraded the frontend compiler from TypeScript 6 to the native (Go) **TypeScript 7.0.2** (GA). No source changes were required — `tsc --noEmit`, the production build (`tsc && vite build`), and the full test suite (123 tests) all pass unchanged. Type-checking is **~5.5× faster** (`tsc --noEmit` ~3.8s → ~0.7s), speeding up both CI's TypeScript Check job and local builds. The native compiler distributes as per-platform binaries via optional dependencies, so the lockfile carries `@typescript/typescript-<os>-<arch>` entries for each CI target.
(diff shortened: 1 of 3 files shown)


adamgell:
Scope note: this PR now also brings `CHANGELOG.md` `[Unreleased]` up to date with the user-facing work merged since 1.3.2 — log error navigation (#192), Teams MSIX log collection (#242), session filter restore + error logging (#244), tail truncation/rotation reset (#243), and the Tauri 2.11/build repair (#240), alongside the TypeScript 7 entry. Dependency bumps, CI-internal changes, and README-only docs were intentionally left out.

PR #248, opened 2026-07-11
feat(parser): treat CCM type="0" as a first-class Success severity

### LINKED ISSUE #211: `type="0"` being picked up as `Info` and not `Success`.
Hey there, super cool project! I'm really glad to see an open source CCM log viewer out in the wild! And a proper contender to replace OneTrace, at that.

In the CCM format, a type of 0 indicates success and is often highlighted as such, however here it's only showing as informational. You'll note the green tick in the status column of OneTrace.

[PSAppDeployToolkit_4.2.0_EN_01_PSAppDeployToolkit_Install.log](https://github.com/user-attachments/files/28630527/PSAppDeployToolkit_4.2.0_EN_01_PSAppDeployToolkit_Install.log)

Image

Image

adamgell:
Fixes the issue reported in #211: in the CCM format `type="0"` indicates a **successful** operation (the green tick in OneTrace's status column), but the parser was only showing it as informational. This adds a first-class `Success` severity and wires it through the whole stack.

Re-implements the intent of #219 (closed) cleanly against current `main` — that branch was based on a pre-workspace-repair tree and carried a large stale `Cargo.lock`/CI/`deny.toml` diff. This PR is the feature only: **12 files, +159/−13**, no lockfile or CI churn.

## Parser (Rust)
- Added a `Success` variant to the `Severity` enum (`0=Success, 1=Info, 2=Warning, 3=Error`).
- `severity_from_type_field`: explicit `type="0"` → `Severity::Success`.
- **Latent bug fixed:** in `parse_captures`, an absent/empty `type=""` was coerced to `Some(0)` via `unwrap_or(0)`. With `0` now meaning Success, that would misclassify neutral lines. It's now preserved as `None` so it falls back to text-based detection (still `Info` for neutral messages). This aligns the CCM path with the CmtLog and IME paths, which already passed `Option<u32>`.
- `filter.rs`: added the `Severity::Success => "Success"` arm so severity filtering works.
- Tests: `type="0"` → Success, the full type-field mapping table, and a guard that an absent type with a neutral message is **not** coerced to Success.

## Frontend (TypeScript)
- `Severity` type gains `"Success"`.
- `LogSeverityPalette` gains `success: { background, text }`; green values added for **all 8 themes**, each aligned with that theme's existing `status.success` color and matching its error/warning contrast.
- `LogRow` renders Success rows and the severity dot in green.
- `useQuickStats` + `QuickStatsPanel` + `StatCard` add a **Success** stat card (Fluent `colorStatusSuccess*` tokens); `StatusBar` counts and shows `N success`.
- New `LogRow.test.tsx` renders a Success entry and asserts it uses the theme's green (`rgb(220, 252, 231)` bg / `rgb(20, 83, 45)` text) and is labeled `Success`, while Info stays neutral.

## Verification
- `cargo test` — full workspace green; new CCM tests pass; the existing empty-`type` test still passes (now Info via text detection rather than coercion).
- `cargo clippy --all-targets -- -D warnings` — clean.
- `npx tsc --noEmit` — clean.
- `npx vitest run` — 135 passed (+3 new).

> Note: the live UI wasn't driven in a browser because the app requires the Tauri backend to render (frontend-only mode throws in `<AppShell>` on the failing `invoke()` calls, and the full Tauri app opens a native window). The `LogRow` render test observes the actual green-row output instead.

Closes #211.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,6 @@
 *.ps1 text eol=crlf
 *.psm1 text eol=crlf
 *.psd1 text eol=crlf
+
+# Real-world log fixtures are byte-sensitive (UTF-8 BOM + CRLF); never normalize them.
+src-tauri/tests/fixtures/** -text
--- a/crates/cmtraceopen-parser/src/models/log_entry.rs
+++ b/crates/cmtraceopen-parser/src/models/log_entry.rs
@@ -1,9 +1,12 @@
 use serde::{Deserialize, Serialize};
 
 /// Log entry severity level.
-/// Maps directly to CMTrace's type field: 1=Info, 2=Warning, 3=Error
+/// Maps directly to CMTrace's type field: 0=Success, 1=Info, 2=Warning, 3=Error.
+/// `Success` corresponds to CCM/PSADT `type="0"` — a completed operation that
+/// OneTrace renders with a green tick.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Severity {
+    Success,
     Info,
     Warning,
     Error,
--- a/crates/cmtraceopen-parser/src/parser/ccm.rs
+++ b/crates/cmtraceopen-parser/src/parser/ccm.rs
@@ -103,11 +103,11 @@ pub(crate) fn build_timestamp(
 
 pub(crate) fn severity_from_type_field(type_value: Option<u32>, message: &str) -> Severity {
     match type_value {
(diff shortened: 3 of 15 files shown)


hiSandog:
Could persisted filters or saved frontend state contain the previous closed set of severity strings? If so, adding `Success` may need a compatibility test showing that older stored filters deserialize without dropping the filter or failing the view initialization. It would also be useful to pin the behavior of an explicit unknown numeric `type` separately from an absent value, so future format extensions do not silently become `Info` or `Success`.

PR #313, opened 2026-07-30
ci(packaging): own winget locale metadata, add Scoop and Homebrew

adamgell:
Three packaging channels, all built so they cannot silently go stale. The third-party Chocolatey package sitting on 1.3.2 while 1.5.0 shipped is the failure mode being designed against.

## 1. winget: own the metadata komac cannot manage

`komac update` copies `PackageUrl`, `ShortDescription`, `Description`, `Moniker`, and `Tags` forward from the previously published manifest and exposes no CLI flag for any of them. All of the following is live today on `AdamGell.CMTraceOpen`, and none of it would ever have self-healed, because every future release inherits it:

* `ShortDescription` is still the pre-1.0 text, "An open-source log viewer inspired by Microsoft's CMTrace.exe."
* `Tags` has 2 entries, so `winget search sccm`, `winget search configmgr`, and `winget search "log viewer"` do not match the package.
* No `Moniker`, so `winget install cmtrace` does not work.
* `Description` was never set at all.

Submission is now generate, patch, submit:

* `.github/winget/locale-metadata.yaml` is the source of truth for those five keys. Adds `Moniker: cmtrace`, expands `Tags` from 2 to 16, adds the absent `Description`, rewrites `ShortDescription`, points `PackageUrl` at cmtraceopen.com.
* `.github/scripts/Set-WingetLocaleMetadata.ps1` splices them into the generated manifest. Replaces keys in place to preserve schema order, inserts absent keys at their schema position, preserves CRLF (a LF rewrite is a whole-file diff that fails winget validation), and fails the build on over-limit values.
* `winget-publish.yml` gains the `actions/checkout` step it never had.

## 2. Homebrew cask (macOS)

* `Casks/cmtrace-open.rb` for the aarch64 DMG. The token is the mechanical derivation from `CMTrace Open.app`; Homebrew's rules rule out owning the bare `cmtrace` token, so a second `name "CMTrace"` stanza gives short-word search matching instead.
* **No workflow, deliberately.** Homebrew's autobump is opt-out, not opt-in: casks in homebrew-cask are bumped by a Homebrew-run action that recomputes the sha256 itself. A local `brew bump-cask-pr` job would need a token with fork push access, would race Homebrew's own action, and would duplicate free work.
* `livecheck` uses `strategy :github_latest`, which hits `releases/latest` and therefore ignores the rolling `nightly` prerelease tag. Verified live: it resolves 1.5.0, not the nightly.
* `zap trash:` paths were enumerated from a machine that had actually run the release build. `Caches/`, `HTTPStorages/`, and `Saved Application State/` were omitted because they were genuinely absent, rather than listed on macOS convention.

Arm64 only. On an Intel Mac, `depends_on arch: :arm64` makes Homebrew fail with a hard error rather than falling back to Rosetta, because there is no x86_64 or universal artifact to translate.

## 3. Scoop bucket (Windows)

* `bucket/cmtrace.json` and `bucket/cmtrace-lite.json` for the portable single-file executables, x64 and arm64. This repo doubles as its own Scoop bucket (`bucket/` is the directory name Scoop recognises).
* `.github/scripts/Update-ScoopManifest.ps1` rewrites version, urls, and hashes from each manifest's own `autoupdate` templates, so this bucket and Scoop's Excavator cannot drift on where assets live. It edits the JSON text in place rather than round-tripping through `ConvertTo-Json`, which reindents to two spaces and mangles the nested array in `shortcuts`. Hashes come from the GitHub asset digest with a download fallback, so it never writes a hash it did not verify, and it re-parses after splicing to fail on any mismatch.
* `.github/workflows/scoop-publish.yml` bumps the manifests on release, then **installs real Scoop on a Windows runner and performs a full install and uninstall of each manifest before committing**, so a bad hash fails the release job instead of every user's `scoop install`.

## Verification

* All four Scoop hashes match the digests GitHub reports for the assets.
* `brew audit --new --cask` exits 0 with zero offenses; `brew style` reports none. Both run through a throwaway tap, because Homebrew refuses a cask outside one.
* `brew livecheck` resolves 1.5.0.
* Both Scoop manifests parse, and their field order matches Scoop's contributing guide.
* `Update-ScoopManifest.ps1` reproduces both manifests byte-identically at 1.5.0, changes exactly 6 lines at 1.4.0, and was exercised on its download-fallback and failure paths.
* `Set-WingetLocaleMetadata.ps1` output is byte-identical to an independently built reference against the real published 1.5.0 manifest, idempotent, with `ReleaseNotes` and all installer-facing keys untouched and CRLF preserved.
* Both PowerShell scripts parse with 0 errors and contain no non-ASCII bytes.
* `npx tsc --noEmit` clean.

## Known gaps, please read before merging

1. **The winget submit path has never run.** The `komac update --dry-run --output-directory` then `komac submit` sequence is derived from komac's argument definitions, not a live run. Validate with a `workflow_dispatch` against `v1.5.0` before 1.6.0 depends on it. The submit step resolves the directory from the generated locale manifest rather than hardcoding a path, which should absorb most layout variation.
2. **`scoop install` has not run on a real Scoop.** No Scoop on the dev host. That is exactly what the Windows runner validation step exists for, but it has not executed yet. A `workflow_dispatch` with `tag: v1.5.0` would confirm it.
3. **The Homebrew install, uninstall, and zap round-trip was deliberately not run.** `brew zap` would have deleted real user data on the dev machine. This needs running somewhere disposable before submission.
4. **`scoop-publish.yml` pushes directly to the default branch.** `main` is currently unprotected, which was verified. If branch protection is ever added, this workflow must switch to opening a PR. There is also no `concurrency` guard and no retry on a non-fast-forward push, so two releases in quick succession could leave the bucket a version behind until the next release or a manual dispatch.
5. **Action pinning is inconsistent across the repo.** `cmtrace-release.yml` pins by SHA while `winget-publish.yml` uses `actions/checkout@v4`. Both new workflows match the latter. Worth normalising separately.

## One-time manual submissions, not done here

* **homebrew-cask**: copy the cask into a fork at `Casks/c/cmtrace-open.rb` and open a hand-written PR. `brew bump-cask-pr` only works on casks that already exist upstream. Steps in `Casks/README.md`.
* **ScoopInstaller/Extras**: open an issue first, as their guide asks, then a PR titled `cmtrace: Add version 1.5.0`. Recommendation in `bucket/README.md` is to submit `cmtrace` alone at first, since a second manifest for the same project raises review friction on a first submission. Both ship in the self-hosted bucket regardless.

No pull requests were opened against any external repository.

## Summary by CodeRabbit

- **New Features**
  - Added Homebrew Cask support for macOS Apple Silicon.
  - Added Scoop manifests for CMTrace Open and CMTrace Lite, including x64/arm64 downloads and automatic updates.
  - Added automated publishing for Scoop and Winget packages.
  - Improved Winget package metadata, including descriptions, tags, and links.

- **Documentation**
  - Added installation instructions for Homebrew and Scoop.
  - Added guidance for using and maintaining the Scoop bucket and Homebrew Cask.

### DIFF
--- a/.github/scripts/Set-WingetLocaleMetadata.ps1
+++ b/.github/scripts/Set-WingetLocaleMetadata.ps1
@@ -1,0 +1,203 @@
+<#
+.SYNOPSIS
+    Applies repo-owned defaultLocale metadata to komac-generated winget manifests.
+
+.DESCRIPTION
+    'komac update' carries PackageUrl, ShortDescription, Description, Moniker and
+    Tags forward from the previously published manifest and offers no CLI flags to
+    change them, so a stale value stays stale for every future release. This script
+    splices the keys from .github/winget/locale-metadata.yaml into the generated
+    defaultLocale manifest, between 'komac update --dry-run --output-directory' and
+    'komac submit'.
+
+    Keys already present in the generated manifest are replaced in place, which
+    preserves schema key order. Keys komac omitted entirely are inserted at their
+    correct schema position. Line endings of the generated manifest are preserved
+    (winget-pkgs manifests are CRLF; rewriting them as LF produces a whole-file
+    diff and fails validation).
+
+.PARAMETER ManifestDirectory
+    Directory holding the komac-generated manifests. Searched recursively for
+    '*.locale.en-US.yaml'.
+
+.PARAMETER MetadataPath
+    Path to the metadata YAML whose top-level keys should be applied.
+
+.EXAMPLE
+    ./.github/scripts/Set-WingetLocaleMetadata.ps1 -ManifestDirectory "$env:RUNNER_TEMP\winget-out" -MetadataPath ./.github/winget/locale-metadata.yaml
+#>
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$ManifestDirectory,
+
+    [Parameter(Mandatory = $true)]
+    [string]$MetadataPath
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+# Key order from the defaultLocale manifest schema. Only used to position keys
+# that the generated manifest does not already contain.
+$schemaOrder = @(
+    'PackageIdentifier', 'PackageVersion', 'PackageLocale', 'Publisher',
+    'PublisherUrl', 'PublisherSupportUrl', 'PrivacyUrl', 'Author', 'PackageName',
+    'PackageUrl', 'License', 'LicenseUrl', 'Copyright', 'CopyrightUrl',
+    'ShortDescription', 'Description', 'Moniker', 'Tags', 'Agreements',
+    'ReleaseNotes', 'ReleaseNotesUrl', 'PurchaseUrl', 'InstallationNotes',
+    'Documentations', 'Icons', 'ManifestType', 'ManifestVersion'
+)
+
+function Get-YamlSegments {
+    <#
+        Splits flat YAML text into ordered segments, one per top-level key. Each
+        segment owns its continuation lines (block scalar bodies, sequence items),
+        so replacing a segment replaces the whole value.
+    #>
+    param([Parameter(Mandatory = $true)][AllowEmptyCollection()][AllowEmptyString()][string[]]$Lines)
+
+    $preamble = [System.Collections.Generic.List[string]]::new()
+    $segments = [System.Collections.Generic.List[object]]::new()
+
+    foreach ($line in $Lines) {
+        if ($line -match '^([A-Za-z][A-Za-z0-9_]*):') {
+            $segment = [pscustomobject]@{
+                Key   = $Matches[1]
+                Lines = [System.Collections.Generic.List[string]]::new()
+            }
+            $segment.Lines.Add($line)
+            $segments.Add($segment)
+        }
+        elseif ($segments.Count -gt 0) {
+            $segments[$segments.Count - 1].Lines.Add($line)
+        }
+        else {
+            $preamble.Add($line)
+        }
+    }
+
+    return [pscustomobject]@{ Preamble = $preamble; Segments = $segments }
+}
+
+function Assert-MetadataLimits {
+    <#
+        Fails the build on values winget would reject, so a bad edit to the
+        metadata file surfaces here rather than as a rejected pull request.
+    #>
+    param([Parameter(Mandatory = $true)][object[]]$Segments)
+
+    foreach ($segment in $Segments) {
+        switch ($segment.Key) {
+            'ShortDescription' {
+                $value = ($segment.Lines[0] -replace '^ShortDescription:\s*', '')
+                if ($value.Length -gt 256) {
+                    throw "ShortDescription is $($value.Length) characters; winget allows 256."
+                }
+            }
+            'Tags' {
+                $tags = @($segment.Lines | Where-Object { $_ -match '^-\s+' })
+                if ($tags.Count -gt 16) {
+                    throw "Tags has $($tags.Count) entries; winget allows 16."
+                }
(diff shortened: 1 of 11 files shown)


adamgell:
## Verification update: two of the three known gaps are now closed, and the rehearsal found a real bug

### winget (gap 1)

The rehearsal is now a real rehearsal. Two changes were needed first:

1. **`workflow_dispatch` would have submitted for real.** `--dry-run` only covers `komac update`; the separate `komac submit` step ran unconditionally, so dispatching against `v1.5.0` (which its own input description invites) would have opened a duplicate pull request on microsoft/winget-pkgs. Submission is now gated on `github.event_name == 'release'`, and a dispatch echoes the patched manifest instead.
2. **The flag was wrong.** The first dispatch failed immediately:

   ```
   error: unexpected argument '--output-directory' found
   tip: a similar argument exists: '--output'
   ```

   The name was inferred from komac's argument definition, `#[arg(short, long, env = "OUTPUT_DIRECTORY", ...)]`. clap derives the long flag from the field name, which is `output`; `OUTPUT_DIRECTORY` is only the environment variable. Fixed in 28e4fde.

Run [30509675175](https://github.com/adamgell/cmtraceopen/actions/runs/30509675175) then passed end to end, with `Submit to winget-pkgs` correctly skipped. The log proves the whole chain:

* komac wrote to a **nested** path, `manifests/a/AdamGell/CMTraceOpen/1.5.0`. Resolving the submit directory from the generated locale manifest rather than hardcoding it was therefore load-bearing.
* Incoming metadata was stale exactly as diagnosed: `PackageUrl: https://github.com/adamgell/cmtraceopen`, `ShortDescription: An open-source log viewer inspired by Microsoft's CMTrace.exe.`, `Tags: [cmtrace, intune]`.
* The patch script reported `replaced PackageUrl`, `replaced ShortDescription`, `inserted Description`, `inserted Moniker`, `replaced Tags`, and `CRLF preserved`.
* Output carries `Moniker: cmtrace` and all 16 alphabetically sorted tags.

**Still unrun:** `komac submit` itself, which by design only fires on a published release.

### Homebrew (gap 3)

The install, uninstall, and zap round-trip has been run. User data was backed up first and restored afterwards, verified by checksum.

* `brew install --cask` places the app.
* `brew uninstall --cask` removes the app and **leaves all four user-data paths intact**.
* `brew uninstall --zap --cask` removes the app and trashes exactly the four paths in the `zap trash:` stanza, no more and no fewer. That confirms the empirically-enumerated stanza is correct.

This also surfaced a documentation bug: **there is no `brew zap` command**, it is `brew uninstall --zap`. `Casks/README.md` documented the non-existent form. Fixed in 28e4fde.

Template-exact reruns for the upstream submission, all exit 0 with no offenses: `brew audit --cask --online`, `brew audit --cask --new`, `brew style --fix` (changed nothing), `HOMEBREW_NO_INSTALL_FROM_API=1 brew install --cask`, `brew uninstall --cask`.

### Upstream submission

Homebrew/homebrew-cask#278064 is open, branched from upstream `main`, with AI use disclosed per their policy. Their bot auto-closed the first attempt for an incomplete template and reopened it once the template was filled in. Their CI (`tap_syntax`, `test cmtrace-open (macos-26, arm)`) is running.

### crates.io

Both crates are published: `cmtraceopen-parser` 0.1.1 (carrying the metadata 0.1.0 shipped without) and `cmtrace-open` 1.5.0. The app crate needed `--no-verify`, as anticipated: its tarball **compiles fine**, but `tauri-build` writes `gen/schemas` outside `OUT_DIR`, which trips cargo's source-modification check after a successful build.

### Remaining gap

Gap 2 is unchanged: `scoop install` has never run against real Scoop. The workflow's Windows validation step exists for exactly that but has not executed, since it only runs on release or dispatch of the Scoop workflow.

Issue #324, opened 2026-07-30
SCCM Client: diagnose task-sequence execution across WinPE and post-client phases


adamgell:
## Parent

Part of #317.

## Outcome

Identify the exact Task Sequence phase and terminal step while preserving the changing SMSTS log location.

## Initial input contract

- SMSTS and TSAgent evidence from WinPE, pre-client, client-installed, and post-completion paths.
- Related content/location, command, setup, reboot, and deployment-output artifacts where supplied.

## Acceptance criteria

- Classify capture phase and log relocation without merging unrelated Task Sequence runs.
- Identify exact failed step, command/exit code, content/DP failure, prerequisite/dependency, or reboot blocker with cited entries.
- Include fixtures for early WinPE, post-format, new OS/pre-client, client-installed, and completed paths.
- Represent missing earlier/later phase evidence as a coverage gap and request it explicitly.
- Preserve existing generic CCM/Panther/servicing parsers as supplements.

## Boundary

This issue does not introduce filesystem discovery into the pure parser.

Microsoft reference: https://learn.microsoft.com/en-us/intune/configmgr/core/plan-design/hierarchy/log-files

## Research-integrated safeguards

- Use path provenance and _SMSTSLogPath evidence when present to classify relocation, and key runs on execution/step context rather than filename or time alone.
- Add fixtures with relocated fragments, unrelated Task Sequence runs in one directory, and missing earlier/later stages.
- Preserve original timestamps/offsets and normalize only for ordering; a missing stage remains a coverage gap, not a synthetic failure.

adamgell:
### Corpus-preparation ownership claimed

I’m starting the #324 Task Sequence corpus-preparation slice in isolated worktree `.worktrees/sccm-324-task-sequence-corpus`, branch `codex/sccm-324-task-sequence-corpus`, from freshly verified live parser-family base `a58eefdc2063f9e03899af882b0bcc35c9cc2bbd`.

Scope is fixtures, a checked-in Rust fixture contract, and preparation documentation only. It will cover the approved WinPE/post-format/pre-client/client-installed/completed-relocated path matrix plus unrelated-run, rotation, incomplete, terminal phase, invalid-offset, unknown-profile, and complete-looking-unkeyed adversarial cases. Production task-sequence reduction/catalog/native capture remains blocked on final reviewed #318/#319 interfaces.

[...]

PR #350, opened 2026-07-30
feat(sccm): add versioned correlation keys

adamgell:
## Scope

Implements Task 6 of the SCCM diagnostic spine for #318:

- adds the public correlation-key, extraction-profile, confidence, and coverage-gap contracts;
- normalizes all 14 declared SCCM lexical key families without adding a parser kind or duplicating CCM parsing;
- gates extraction by ConfigMgr version/profile maturity;
- keeps the single exercised profile experimental and caps every extracted key at Low confidence;
- rejects forged Stable profiles and malformed or prefix-collision version labels;
- preserves deterministic ordered candidate gaps for missing and unvalidated versions;
- requires a declared whole-token boundary both before a label and after its captured value.

No native collection, Windows I/O, findings/reducers, cross-side causal logic, `ParserKind::Sccm`, CCM grammar, `LogEntry`, or dependency changes are included.

## Review corrections

Independent review first found that a regex capture could stop before an unrecognized contiguous value suffix. Values such as `{GUID}}`, `LAB00001é`, or `hosté` could therefore emit a truncated Low key. The first correction made end-of-message, whitespace, comma, semicolon, or ampersand the only accepted value boundaries.

A fresh dynamic rereview then found the symmetric label-start flaw: per-label regex scans treated forbidden punctuation as a word boundary, so `package id=LAB00001/content id=ContentABC` could emit a nested ContentId even though the package token was malformed. Commit `e99cbab924da49d1b296459758307ba1ee8b110c` now applies the same explicit accepted-boundary contract to label starts. It does not special-case `/`, `:`, or `+`.

The test-first nested-label matrix covers all 14 declared second-label kinds across forbidden `/`, `:`, and `+` delimiters, plus Unicode and doubled-brace adjacency. It asserts zero leaked keys, one deterministic cited malformed PackageId gap, UTF-16 spans, and unvalidated-profile candidate behavior. The positive matrix pins accepted whitespace, newline, tab, comma, semicolon, and ampersand starts while retaining Low confidence.

## Restack evidence

Based on live `codex/parser-family-skeleton` at `a58eefdc2063f9e03899af882b0bcc35c9cc2bbd`.

```text
1:  79ca233 = 1:  4bb5473 feat(sccm): add versioned correlation keys
2:  7dd577f = 2:  c808848 fix(sccm): enforce correlation key boundaries
3:  -------- > 3:  e99cbab fix(sccm): reject nested correlation labels
```

[...]

PR #352, opened 2026-07-30
test(sccm): prepare Task Sequence corpus

adamgell:
## Summary

- add a 17-scenario, 22-artifact synthetic client Task Sequence corpus covering WinPE, post-format, pre-client, client-installed, completed/relocated fragments, same-time unrelated runs, rotation boundaries, incomplete coverage, terminal failures, reboot continuation, invalid offsets, unknown profiles, and complete-looking unkeyed evidence
- add a validator-backed Rust contract for exact execution keys, `_SMSTSLogPath` provenance, path classes/relocation order, CCM logical completeness, timestamp provenance, phase/terminal semantics, bounded coverage gaps, deterministic bytes/lines/digest, privacy, and fail-closed adversarial mutations
- require all four declared execution-key fields to co-occur in every cited complete CCM record; aggregate substring pooling across unrelated records is rejected
- derive physical completeness, source-path/profile provenance, relocation order, terminal outcomes, confidence ceilings, finding semantics, and absent-coverage behavior from cited evidence instead of mutually editable JSON claims
- model the split rotation record only as an explicit ordered lo→current logical reconstruction; per-artifact path claims remain physical, and shared fingerprint alone supplies no evidence
- add the issue-scoped preparation specification with supported path/profile scope, unsupported boot/recovery variants, native-validation deferral, and replay gates

## Boundaries

- fixtures, tests, and specification only
- no production Task Sequence reducer, catalog mutation, or native capture
- does not consume the experimental low-confidence production key extractor merged in #350; exact fixture keys remain explicit corpus contracts pending final #318/#319 interfaces
- filename, path, timestamp, display name, and path fingerprint never create or merge execution identity
- missing `smsts` is coverage, not proof that no Task Sequence ran
- missing per-fragment `_SMSTSLogPath` is explicit absence, not permission to borrow another artifact’s observation
- no application, policy, distribution-point, management-point, or server-causality claim
- no native Windows or live ConfigMgr acceptance claim

## Corpus inventory

- 17 scenarios
- 22 artifact rows: 21 captured and 1 absent
- 21 evidence files: 19 complete logical CCM records and 2 partial rotation fragments
- 17 logical coverage rows: 15 captured, 1 partial, 1 absent
- 8,243 bytes, 21 lines
- SHA-256 `917df82bdf96ae4debd3e02e669669a9b564e932d7052091fb39094305593c8b`

## Verification

- focused initial TDD RED: missing Task Sequence fixture root
- focused CodeRabbit blocker RED: a mixed run-A/run-B key was incorrectly accepted when substrings were pooled across citations
- focused independent-review RED at `b880179`: the real `validate_contract` accepted exactly 14/14 coherent path/profile/completeness/order/outcome/confidence/finding/coverage mutations
- focused fresh-review RED at `6f5f2d2`: the real validator accepted exactly 2/2 shared-fingerprint path-provenance mutations
- `cargo test -p cmtraceopen-parser --test sccm_client_task_sequence_fixture_contract` (10/10)
- `cargo test -p cmtraceopen-parser` (664/664)
- `cargo check -p cmtraceopen-parser --target wasm32-unknown-unknown`
- `npx tsc --noEmit`
- scoped `rustfmt --edition 2021 --check`
- `git diff --check`
- independent exact corpus recomputation: 17 scenarios / 22 artifacts / 21 files / 8,243 bytes / 21 lines / 19 complete physical fragments / 2 partial fragments / unchanged digest
- committed-tree proof: all 21 ignored `smsts.log`/`smsts.lo_` evidence files are tracked

## Review evidence

- branch remains based directly on `25b37333affde22b4ef8a19f4a5f3d89c082b599`
- initial preparation commit range-diff: `cddba36 = 94fd08d`
- staged CodeRabbit findings for capture-state wording and escaped-path privacy detection were fixed in `94fd08d`
- exact-head CodeRabbit finding `5137395468` identified cross-record key pooling; the precise adversarial mutation was captured RED and fixed in `b880179199c89a5b7b67686e7fe0e1a069dfc0a2`
- independent exact-range review `4824185973` reproduced 14 coherent validator bypasses against the real private validator; all 14 are committed as fail-closed mutations and fixed in `6f5f2d2e4341feace86e422008f397fd519bd353`
- fresh CodeRabbit exact-head review `5137635021` found shared-fingerprint path borrowing; changed-token and same-fingerprint-donor mutations were captured RED and fixed in `37dc6d08920c9dcb04a325f5dfe7531531695d9a`
- the three accepted inline review threads were replied to and resolved only after their mutations were green
- fresh exact-head CodeRabbit response `5137740173` re-reviewed `37dc6d08920c9dcb04a325f5dfe7531531695d9a` against direct base `25b37333affde22b4ef8a19f4a5f3d89c082b599` and reported no findings; the PR remains draft
- the next independent review is intentionally deferred until #318 merges and this draft branch is restacked onto the resulting live base

Refs #324

Epic: #317

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for SCCM client task-sequence scenarios, including successful, failed, incomplete, rotated, relocated, and ambiguous logs.
  * Added validation for artifact provenance, ordering, coverage, timestamps, privacy boundaries, correlation rules, and conservative classifications.
  * Added integrity checks across a broad synthetic scenario corpus.

* **Documentation**
  * Documented the synthetic task-sequence corpus, supported scenarios, validation rules, integrity checks, privacy constraints, and replay guidance.


adamgell:
@coderabbitai review

Please review exact head `94fd08d3e5e1f7024d6fc78335d2064c96639852` against base `25b37333affde22b4ef8a19f4a5f3d89c082b599`.

Scope is fixtures/tests/specification only. Treat the explicit synthetic execution keys as corpus contracts; do not infer production reducer readiness from the experimental low-confidence extractor. Please focus on false joins, physical-vs-logical coverage, relocation/rotation ordering, CCM record completeness, phase/terminal semantics, privacy, determinism, and native/server-causality overclaiming. The two findings from the staged review were fixed before this exact-head request.

adamgell:
@coderabbitai review

Please re-review exact head `b880179199c89a5b7b67686e7fe0e1a069dfc0a2` against base `25b37333affde22b4ef8a19f4a5f3d89c082b599`.

Tracking finding https://github.com/adamgell/cmtraceopen/pull/352#issuecomment-5137395468: fixed in `b880179`.

- captured the precise same-time `unrelated-runs` mixed-key mutation RED: run A `executionId`/`runContext` plus run B `advertisementId` was previously accepted through aggregate substring pooling
- every transaction evidence reference must now resolve to one complete normalized CCM record containing every declared non-metadata key field in that same record
- phase/state must co-occur in one already key-qualified cited record
- path-sequence, ordering, and terminal references must come from the key-bound transaction evidence set
- focused contract 8/8, full parser 662/662, clippy, wasm32, TypeScript, scoped rustfmt, and diff checks are green

Please verify false-join closure and the existing preparation-only/no-production-readiness boundaries. A status context alone is not being treated as substantive approval.

adamgell (commented):
## Exact-range independent adversarial rereview — BLOCKED

Reviewed exact base `25b37333affde22b4ef8a19f4a5f3d89c082b599` through exact head `b880179199c89a5b7b67686e7fe0e1a069dfc0a2` in a clean detached worktree. The committed corpus is internally consistent, and the per-record execution-key fix is good, but the claimed fail-closed validator contract is not yet sound enough to merge.

I compiled the committed test file unchanged into an external review harness with `include!`, pointed it at the exact committed fixtures, mutated the in-memory `manifest.json`/`expected.json` values, and invoked the real private `validate_contract` function. The baseline probes correctly rejected:

- a four-field key split across two unrelated cited records;
- each independently swapped `executionId`, `taskSequencePackageId`, `advertisementId`, and `runContext`;
- same-time unrelated record pooling;
- duplicate physical/canonical aliases;
- one-field path/profile/timestamp/offset/coverage/relocation/rotation drift;
- direct unkeyed confidence escalation;
- missing failure terminal evidence, invalid phase, unbounded transaction request, and absent-to-captured coverage.

However, the same real validator returned `Ok(())` for **14 coherent adversarial mutations**:

1. coherent `pathClass` drift;
2. coherent `_SMSTSLogPath`/sanitized-path drift away from the cited record;
3. `sourceVersion` drift;
4. declared extraction-profile drift;
5. invalid capture-timestamp drift;
6. an incomplete `smsts.lo_` fragment promoted to complete logical coverage;
7. coherent relocation-order drift;
8. complete-looking unkeyed evidence promoted to a success finding;
9. a succeeded transaction with no terminal citation;
10. a nonterminal finding promoted to `confirmedFailure`;
11. an invalid-offset transaction promoted to high confidence;
12. an unknown-version source promoted to an exact transaction under the synthetic profile;
13. an unbounded finding-level next-artifact request; and
14. an absent source promoted to a success/no-run finding.

These are not production-parser concerns; they are holes in the new fixture validator itself. The preparation spec says source/profile provenance is authoritative, unknown profiles stay source-local/Low, `Complete` needs terminal evidence, partial physical fragments cannot become logical records, findings remain evidence-first, and absence is coverage only. A future fixture change can currently violate those contracts while the advertised validator still passes.

### Reconciliation and gates

- Independent inventory: 17 scenarios; 22 artifacts (21 captured, 1 absent); 21 physical files; 8,243 bytes; 21 lines; 19 complete flags and 2 partial flags; digest `917df82bdf96ae4debd3e02e669669a9b564e932d7052091fb39094305593c8b`.
- `git ls-files --stage` contains all 21 physical `smsts.log`/`smsts.lo_` files. All 17 manifests and 21 physical files carry explicit synthetic markers/handles. The privacy scan covers plain and JSON-escaped Windows-path forms, SIDs, email-like identities, credentials/tokens, and prohibited native/server/app-policy claims; no fixture hit.
- Exact diff is fixtures, one test target, and the preparation spec only—no production reducer/catalog/native code or dependency changes.
- Focused target: 8/8 passed.
- Full parser: 662/662 passed.
- Strict Clippy, wasm32 check, TypeScript, scoped Rust 1.88 formatting, and exact-range `git diff --check`: passed.
- External adversarial harness: 2/2 reviewer tests passed; one proves the expected fail-closed edges, while the other records the 14 validator bypasses above.

The fresh exact-head [CodeRabbit GitHub response](https://github.com/adamgell/cmtraceopen/pull/352#issuecomment-5137434953) reported no findings. A second exact-range CodeRabbit CLI attempt was rate-limited; I did not treat the rate status as approval. The executable mutation evidence above is the blocking result.

This review is anchored to `b880179199c89a5b7b67686e7fe0e1a069dfc0a2`. It does not mark the draft ready, merge it, close #324, or claim native/live acceptance.

crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs
```diff
@@ -0,0 +1,1750 @@
...
+            if !sanitized_path.starts_with("SYNTHETIC://")
+                || artifact["smstsLogPathEvidence"] != sanitized_path
+            {
+                return Err(format!(
+                    "{scenario}/{artifact_id}: _SMSTSLogPath provenance is not bound"
+                ));
+            }
+            let contents = std::fs::read_to_string(&fixture_path)
+                .map_err(|error| format!("{relative_path} is not UTF-8: {error}"))?;
+            if artifact["rotation"]["fragmentComplete"] == true
```
adamgell: [P1] Derive logical completeness from the CCM grammar before trusting `fragmentComplete`.

The storage validator treats this Boolean plus a synthetic marker as proof of completeness. I set only the archived `smsts.lo_` prefix's `fragmentComplete` to true and coherently updated expected coverage/provenance; the real validator accepted `captured` logical coverage even though that physical file still normalizes to zero CCM records. The separate fixed-fixture test notices this today, but a coherent fixture edit can update the flag and retain a green validator. Require a `fragmentComplete: true` artifact to normalize to at least one complete CCM record (and reject parse errors); keep controlled `lo`+`current` reconstruction explicitly partial until its reviewed join contract exists.

crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs
```diff
@@ -0,0 +1,1750 @@
...
+        let state = artifact["captureState"]
+            .as_str()
+            .ok_or_else(|| format!("{scenario}/{artifact_id}: captureState is missing"))?;
+        if state == "captured" {
+            if artifact["encoding"] != "utf-8" {
+                return Err(format!("{scenario}/{artifact_id}: captured encoding"));
+            }
+            if artifact["collectionLimit"]["byteLimit"] != 4096
+                || artifact["collectionLimit"]["limitApplied"] != false
+                || !artifact["sourceVersion"].is_string()
```
adamgell, line 703: [P1] Bind the selected profile and path provenance to the cited record, not just mutually editable JSON fields.

The validator only requires `sourceVersion`/`capturedUtc` to be strings and only compares `sanitizedSourcePath` with the manifest's `smstsLogPathEvidence`. In an executable harness using this exact `validate_contract`, I coherently changed `pathClass`, both path fields, expected provenance, and the declared extraction profile; I also changed `sourceVersion` to `5.00.UNKNOWN.0000` and `capturedUtc` to `not-a-timestamp`. Every mutation was accepted. Most critically, the unknown-version fixture can be changed from a Low source-local candidate into an exact transaction under `task-sequence-client-5.00.test-v1`, because the transaction's hard-coded profile string is never joined back to the artifact version or top-level selected profile. Parse and pin capture time, bind the reviewed profile to allowed source versions/path classes, and compare `_SMSTSLogPath` against the value in each cited complete record.

crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs
```diff
@@ -0,0 +1,1750 @@
...
+        evidence_text(scenario_root, &artifacts_by_id, &observation["evidence"])?;
+    }
+
+    let finding_ids = sorted_ids(&expected["findings"], "findingId");
+    let mut sorted_finding_ids = finding_ids.clone();
+    sorted_finding_ids.sort();
+    if finding_ids != sorted_finding_ids {
+        return Err(format!("{scenario}: finding IDs are not sorted"));
+    }
+    for finding in expected["findings"]
```
adamgell, line 1790: [P1] Validate terminal/confidence/finding semantics instead of only checking that a finding has a citation or coverage row.

The executable mutation harness showed this block accepts: complete-looking unkeyed evidence reclassified as success after deleting its Low observation; a nonterminal record reclassified as `confirmedFailure`; an absent source reclassified as success; an invalid-offset transaction raised to High; and an arbitrary finding-level `boundedNextArtifact` (`entire-device`/`all`). The transaction check above also requires terminal evidence only for `confirmedFailure`, so a `succeeded`/`complete` transaction with `terminalEvidence: null` passes. Enforce the allowed classification/state/confidence matrix, require same-key terminal evidence for both confirmed failure and success/Complete, cap unusable offsets and unknown/unkeyed profiles, validate finding evidence terminality, and apply the same bounded logical-source/path-class rule to finding requests.

crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs
adamgell: Fixed at exact head `6f5f2d2e4341feace86e422008f397fd519bd353`. Capture timestamps must parse as RFC 3339; sanitized `_SMSTSLogPath` and `pathClass` are derived from physical evidence (including the controlled same-fingerprint rotation pair); and the selected extraction profile is derived from recognized sourceVersion/pathClass evidence and must match every exact key. The committed coherent-mutation test now rejects path/pathClass, version/profile, timestamp, and unknown-version exact/High promotions. Focused contract target: 9/9 green.

crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs
adamgell: Fixed at exact head `6f5f2d2e4341feace86e422008f397fd519bd353`. `fragmentComplete` is now derived fail-closed from the real CCM grammar: zero parse errors, at least one normalized complete record, and a CCM entry are required; partial `smsts.lo_`/current fragments remain partial. The committed coherent-mutation test reproduces this bypass against `validate_contract` and now rejects it. Focused contract target: 9/9 green.

crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs
adamgell: Fixed at exact head `6f5f2d2e4341feace86e422008f397fd519bd353`. The validator now enforces the classification/state/confidence matrix, Low ceiling for non-normalized timestamps, same-key terminal citations for success and confirmed failure, transaction-bound terminal/deferred finding outcomes, bounded finding requests, and coverage-gap/no-run fail-closed behavior. Evidence-derived relocation order is also pinned. All corresponding coherent mutations are committed and rejected; focused contract target is 9/9 green.

adamgell:
@coderabbitai review

Please perform a fresh substantive exact-head review of `6f5f2d2e4341feace86e422008f397fd519bd353` against direct base `25b37333affde22b4ef8a19f4a5f3d89c082b599`.

This follow-up closes the 14 coherent validator bypasses independently reproduced at `b880179`. Please inspect the committed real-`validate_contract` mutation test and re-evaluate each closure rather than relying only on incremental diff summaries: physical CCM completeness; `_SMSTSLogPath`/pathClass provenance; sourceVersion/profile/capture-time binding; evidence-derived relocation order; terminal success/failure citations; classification/state/confidence ceilings; unknown/unkeyed and invalid-offset promotion; finding terminality/request bounds; and absent-coverage/no-run behavior. Preserve the existing per-record exact-key, physical inventory/digest, synthetic privacy, preparation-only, and no-native/no-production boundaries.

Local exact-head proof: focused 9/9, full parser 663/663, strict Clippy, wasm32, TypeScript, scoped Rustfmt, and diff checks pass; corpus digest remains `917df82bdf96ae4debd3e02e669669a9b564e932d7052091fb39094305593c8b`. PR remains draft.

adamgell:
@coderabbitai review

Please re-review exact head `37dc6d08920c9dcb04a325f5dfe7531531695d9a` against direct base `25b37333affde22b4ef8a19f4a5f3d89c082b599`.

Tracking P2: https://github.com/adamgell/cmtraceopen/pull/352#issuecomment-5137635021. Fixed in `37dc6d0` with a test-first closure against the real `validate_contract`:

- RED at `6f5f2d2`: exactly 2/2 mutations were accepted — the current suffix borrowed a coherently changed lo path, and a same-fingerprint donor supplied path provenance after the lo token was removed
- the current suffix now declares per-artifact `smstsLogPathEvidence: null`; its `sanitizedSourcePath` remains capture provenance only
- every physical path claim must be present in that same artifact; complete records require their own observation, while incomplete fragments may explicitly declare absence
- the only allowed logical path reconstruction explicitly orders lo then current, cites the lo line that physically contains `_SMSTSLogPath`, validates both fragments and exact joined CCM grammar, and stays partial/Low/source-local/noncorrelatable
- shared fingerprint alone supplies no evidence; changed-token and same-fingerprint-donor mutations are committed and rejected
- no evidence bytes changed; inventory remains 17 scenarios / 22 artifacts / 21 files / 8,243 bytes / 21 lines / 19 complete / 2 partial, digest `917df82bdf96ae4debd3e02e669669a9b564e932d7052091fb39094305593c8b`
- exact-head gates: focused 10/10, full parser 664/664, strict Clippy, wasm32, TypeScript, scoped Rustfmt, and diff checks pass

Please verify this P2 closure plus the previous 14 coherent-mutation closures and preparation-only/no-production/no-native boundaries. PR remains draft.

adamgell:
Coordinator restack owner active in isolated worktree `.worktrees/sccm-324-restack-a8a`, branch `codex/sccm-324-restack-a8a`, exact preparation head `37dc6d08920c9dcb04a325f5dfe7531531695d9a`. Scope is only merging stable foundation `a8a367c661cd17bd1e73b87724c143ec8565ffc7`, combined-tree verification, and a non-force update after remote-head validation. No production reducer/native/live claim or speculative interface change. Fresh full-range CodeRabbit and a different exact-head reviewer remain mandatory; PR stays draft/open.

adamgell:
🟢 #324 restack is pushed at exact head `823e6878adc1ec2f1a790ae2e3fbe8bc9c09c958`.

Scope/dependency state:
- Merge-only restack of the reviewed #324 task-sequence corpus onto merged #318 foundation `a8a367c661cd17bd1e73b87724c143ec8565ffc7`.
- Merge parents are the prior #324 head `37dc6d08920c9dcb04a325f5dfe7531531695d9a` and foundation `a8a367c661cd17bd1e73b87724c143ec8565ffc7`.
- PR remains draft/open. No merge is requested until a fresh substantive exact-head CodeRabbit review and a different independent exact-head review both pass.

Verification at `823e6878`:
- `cargo +1.88.0 test --locked -p cmtraceopen-parser --test sccm_client_task_sequence_fixture_contract` — 10 passed.
- `cargo +1.88.0 test --locked -p cmtraceopen-parser --test sccm_spine_contract` — 136 passed.
- `cargo +1.88.0 test --locked -p cmtraceopen-parser --test sccm_client_intake_fixture_contract` — 3 passed.
- `cargo +1.88.0 test --locked -p cmtraceopen-parser` — 739 passed, 0 failed.
- `cargo clippy --locked -p cmtraceopen-parser --all-targets --all-features -- -D warnings` — passed.
- `cargo +1.88.0 check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown` — passed.
- `npx tsc --noEmit` — passed during this exact restack validation.
- all changed JSON parsed successfully.
- `rustfmt +1.88.0 --edition 2021 --check crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs` — passed.
- `git diff --check a8a367c6...HEAD` — passed.
- Workspace-wide `cargo +1.88.0 fmt --all -- --check` still reports only inherited formatting drift outside the #324 file set; no #324-owned file is implicated.

Next gate: full-range CodeRabbit at this exact SHA, then independent exact-head review.

adamgell:
@coderabbitai full review

Please perform a fresh full-range review of exact head `823e6878adc1ec2f1a790ae2e3fbe8bc9c09c958` against base `a8a367c661cd17bd1e73b87724c143ec8565ffc7`. Prior review of `37dc6d08` is stale after the foundation restack. Please specifically re-check task-sequence rotation/path provenance, coverage-state semantics, deterministic fixture projection, citation bounds, privacy, and conservative confidence. A status-only, skipped, rate-limited, or stale-SHA response will not be treated as approval.

adamgell:
@coderabbitai review

Please perform a fresh substantive full-range review of exact head 823e6878adc1ec2f1a790ae2e3fbe8bc9c09c958 against foundation a8a367c661cd17bd1e73b87724c143ec8565ffc7. The prior invocation ended under fair-use limits and is not approval. Review Task Sequence phase/path relocation provenance, exact execution keys, rotation/collision/time handling, coverage states, incomplete bundles, deterministic citations, synthetic privacy, parser purity, wasm compatibility, and native-validation disclaimers.

adamgell (commented):
## Exact-SHA independent adversarial review — BLOCK

Reviewed exact base `a8a367c661cd17bd1e73b87724c143ec8565ffc7` through exact head `823e6878adc1ec2f1a790ae2e3fbe8bc9c09c958` in a clean detached worktree. Local HEAD, the remote branch, and PR head all matched before this verdict. I made no repository edits and did not push.

### Blocking contract gaps

1. **P2 — `lastSuccessfulPhase` is not derived from evidence.** The validator checks only that the value belongs to `STATE_CHAIN`; outside the successful `complete` case, it never relates the claimed last-success phase to the cited phase/state evidence. An external harness compiled the exact committed test file with `include!` and invoked the real private `validate_contract`. It accepted all three coherent false histories below:
   - `terminal-preflight`: `start` -> `complete` while the cited transaction is a terminal `preflight` failure;
   - `client-installed`: `setupWindows` -> `complete` while the cited transaction is nonterminal `installClient`;
   - `reboot-continuation`: `installSoftware` -> `complete` while the cited transaction is deferred `postAction`.

   This permits a future fixture/output to claim an uncited later success and violates the monotonic, evidence-first phase contract. Derive the value from cited profile evidence, or enforce the exact admissible predecessor/current-phase rule, and add fail-closed mutations for terminal, in-progress, and deferred cases. This independently reproduces the fresh exact-head CodeRabbit P2.

2. **P2 — the declared exact key kind remains editable.** Changing `key.keyProfileKind` from `executionPackageAdvertisementContext` to `filenameTimestamp` still returns `Ok(())`. Require the field and bind it to the reviewed Task Sequence key profile; an exact four-field identity must not advertise an unrelated/forbidden key kind.

3. **P2 — transaction evidence order is not pinned despite the deterministic-output claim.** Swapping the first and last `relocated-fragments.transactions[0].evidence` references while leaving the phase/path sequence unchanged still returns `Ok(())`. Define and validate canonical evidence ordering (and uniqueness) so `reorderedInputDeterministic: true` is an executable contract rather than an editable assertion.

### Verified areas and gates

- The direct-base diff is preparation-only: synthetic fixtures, one Rust contract target, and one specification; no production reducer, catalog/native code, dependency, network, database, Windows-I/O, or parser-kind change.
- Focused committed target: **10/10 passed**.
- External exact-file harness: **14/14 passed**, including the three reproductions above and a coverage-state matrix proving `absent`, `accessDenied`, `capped`, `skipped`, `unsupported`, `parseFailed`, and `unsafePath` remain `insufficientEvidence`/coverage-only with no server, app/policy, or native-acceptance claim.
- Full parser suite: **739/739 passed**.
- `cargo check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown`: passed on Rust 1.88.
- All 34 fixture JSON files parse; all 21 evidence files are tracked; privacy/overclaim scan, owned-test Rust formatting, exact-range `git diff --check`, and clean-worktree checks passed.
- The exact strict Clippy command is red on 59 pre-existing `clippy::uninlined_format_args` findings in parser source unchanged by this PR. The issue-scoped target passed with only that inherited lint suppressed. Workspace `cargo fmt --check --all` is likewise red only in files outside this PR; the owned Rust target is clean.

The corpus correctly keeps same-time different-key runs separate, treats split rotation as partial/source-local, binds `_SMSTSLogPath` to physical evidence, keeps invalid offsets and unknown profiles conservative, requests bounded artifacts for incomplete coverage, and makes no default-path or native/live acceptance claim. Those strengths do not close the three executable contract holes above.

This BLOCK is anchored only to `823e6878adc1ec2f1a790ae2e3fbe8bc9c09c958`. It does not mark the draft ready, merge it, close #324, or claim native Windows acceptance.

adamgell:
🟢 **#324 correction owner active — exact head `823e6878`**

Isolated ownership:
- worktree: `/Users/Adam.Gell/repo/cmtraceopen/.worktrees/sccm-324-review-fixes-823e`
- local branch: `codex/sccm-324-review-fixes-823e`
- only owned code/test file: `crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs`
- no fixture, production parser, native, dependency, or other agent worktree edits are planned.

TDD-owned regressions:
1. terminal-preflight, client-installed, and reboot-continuation must reject uncited/unobserved `lastSuccessfulPhase` claims;
2. exact Task Sequence keys must reject non-workflow `keyProfileKind` values such as `filenameTimestamp`;
3. relocated transaction evidence references must reject noncanonical ordering.

I will commit the focused RED tests separately, then the smallest validator correction, run the prescribed parser/wasm/Clippy/JSON/privacy/scoped-format/diff gates, guard the remote SHA, push to the existing PR branch, update #324/#317, and request exact-head CodeRabbit. No native/live Windows acceptance will be claimed.

adamgell:
🟢 **#324 review-fix slice pushed — exact head `c8e597e25d89e9e8c390748f90faa1c2cfa853d5`**

Scope remains preparation-only and issue-scoped: only `crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs` changed. No fixture, production reducer, catalog/native adapter, dependency, parser-kind, Windows-I/O, or public model change was made.

TDD commits:
- RED `43cb1f97`: added real-validator regressions for false `lastSuccessfulPhase` histories in terminal/in-progress/deferred transactions, unrelated `keyProfileKind`, and reordered relocated evidence. Focused result was 10 pass / exactly 3 fail.
- GREEN `c8e597e2`: enforces the admissible predecessor/current-phase rule, binds `keyProfileKind` to `executionPackageAdvertisementContext`, and requires transaction evidence to be unique and match the canonical evidence-derived path sequence.

Verification:
- `cargo +1.88.0 test --locked -p cmtraceopen-parser --test sccm_client_task_sequence_fixture_contract` — 13/13 passed.
- `cargo +1.88.0 test --locked -p cmtraceopen-parser` — 742/742 passed.
- `cargo +1.88.0 check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown` — passed.
- `cargo +1.88.0 clippy --locked -p cmtraceopen-parser --test sccm_client_task_sequence_fixture_contract -- -D warnings -A clippy::uninlined_format_args` — passed. The suppression isolates the 59 inherited findings recorded by the exact-head review; this slice does not alter those files.
- All 34 task-sequence JSON files parse; privacy/overclaim scan is clean.
- Owned-file Rust formatting, `git diff --check a8a367c661cd17bd1e73b87724c143ec8565ffc7 HEAD`, and clean-worktree checks passed.

Dependency state: #318 is closed; #319 remains open, so this lane remains a corpus/contract gate and does not add a production reducer. PR remains draft. No native Windows or live SCCM acceptance is claimed.

adamgell:
@coderabbitai review

Please perform a substantive full-range review of exact head `c8e597e25d89e9e8c390748f90faa1c2cfa853d5` against base `a8a367c661cd17bd1e73b87724c143ec8565ffc7`.

Please specifically verify that:
1. terminal-preflight, client-installed, and reboot-continuation cannot claim an uncited later `lastSuccessfulPhase`;
2. exact Task Sequence keys require `keyProfileKind: executionPackageAdvertisementContext` and reject `filenameTimestamp`;
3. relocated transaction evidence is unique and in canonical evidence-derived path order.

Please report substantive findings or an explicit no-findings result. The PR remains draft and makes no native Windows acceptance claim.

adamgell:
🟢 **#324 ownership expansion for CodeRabbit follow-up**

Local CodeRabbit CLI 0.7.1 completed a full committed-range review of exact head `c8e597e25d89e9e8c390748f90faa1c2cfa853d5` against `a8a367c661cd17bd1e73b87724c143ec8565ffc7` and returned five findings. No other #324 agent is active.

This correction lane now explicitly owns only these two files for the follow-up:
- `crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs`
- `docs/sccm/preparation/issue-324-client-task-sequence-corpus.md`

Planned scope: TDD duplicate observation/finding identity rejection and complete-record/null `_SMSTSLogPath` independence in the contract test; minimal specification corrections for bounded path-class terminology, physical-capture versus logical-coverage semantics, and cross-platform digest byte rules. No fixture, production, public model, native adapter, dependency, or parser-kind file will change.

adamgell:
🟢 **CodeRabbit follow-up pushed — exact head `c33669d110ef46e995a9fb2537ba7352ba659796`**

Local CodeRabbit CLI 0.7.1 reviewed the full committed range and returned five findings. All five were verified and addressed within the explicitly claimed #324 Rust contract test and specification; no fixture or production/native file changed.

Follow-up TDD commits:
- RED `6a3b8c0e`: 13 existing tests passed and exactly three new regressions failed for duplicate `observationId`, duplicate `findingId`, and a complete WinPE CCM record with explicit null `_SMSTSLogPath` evidence.
- GREEN `c33669d1`: observation/finding IDs are unique as well as sorted; complete records may declare null only when the physical record truly has no `_SMSTSLogPath`; incomplete missing-path fragments still require explicit controlled reconstruction. The spec now distinguishes logical artifact ID from path class, physical capture from logical completeness/coverage, and defines cross-platform digest bytes and LF behavior.

Current verification:
- focused Task Sequence contract: 16/16 passed;
- full parser suite: 745/745 passed;
- wasm32 parser check: passed;
- issue-scoped strict Clippy with only the recorded inherited `uninlined_format_args` lint suppressed: passed;
- 34 JSON files parse; privacy/overclaim scan clean; all evidence uses LF;
- owned Rust formatting, exact-range diff check, and clean worktree: passed.

PR remains draft. #319 remains open, so no production reducer is added. No native Windows/live SCCM acceptance is claimed.

adamgell:
@coderabbitai review

Please perform a substantive full-range re-review of exact head `c33669d110ef46e995a9fb2537ba7352ba659796` against base `a8a367c661cd17bd1e73b87724c143ec8565ffc7`.

In addition to the original three blockers, please verify the local CodeRabbit follow-ups: unique observation/finding IDs; independence of complete logical CCM framing from absent `_SMSTSLogPath`; bounded path-class terminology; physical-capture versus logical-coverage wording; and canonical cross-platform digest bytes/line endings.

Please report substantive findings or an explicit no-findings result. PR remains draft and makes no native Windows acceptance claim.

adamgell:
🟢 **Second CodeRabbit follow-up pushed — exact head `6629463876c0b08286f2f84deacee0130736dbb8`.**

The second local CodeRabbit full-range pass returned three minor findings, all verified in the owned contract test:
- RED `7c567a4ed6f42c684acf1dbd2320a95d83754b53`: 16 existing tests passed and exactly two temporal regressions failed for equal relocation timestamps and subsecond timestamp truncation.
- GREEN `6629463876c0b08286f2f84deacee0130736dbb8`: equal normalized timestamps are now ambiguous instead of artifact-ID tie-broken; RFC 3339 comparison retains observed subsecond precision; inventory counts scenarios actually traversed.

Verification after the fix: focused 18/18, full parser 747/747, wasm32 check, and issue-scoped strict Clippy passed. The 34 JSON, privacy/overclaim, LF, scoped-format, exact-range diff, and clean-worktree gates also passed before the guarded push.

PR remains draft; #319 remains the production dependency; no native Windows/live SCCM acceptance is claimed.

adamgell:
@coderabbitai review

Please perform a substantive full-range re-review of exact head `6629463876c0b08286f2f84deacee0130736dbb8` against exact base `a8a367c661cd17bd1e73b87724c143ec8565ffc7`.

Please include the latest fixes: equal normalized relocation timestamps must be ambiguous, subsecond evidence must not match a whole-second declaration, and corpus inventory must derive its scenario count from actual traversal. Also recheck all earlier blockers/follow-ups. Report substantive findings or an explicit no-findings result. PR remains draft; no native Windows acceptance claim.

adamgell:
🟢 **Review state at exact head `6629463876c0b08286f2f84deacee0130736dbb8`:** all findings from two completed local CodeRabbit passes are fixed and verified. A third local clean-pass was requested but CodeRabbit CLI returned its recoverable included-review rate limit with a 19-minute retry window. No code or test gate is blocked. The hosted exact-head `@coderabbitai review` request remains active; the draft stays open pending that clean review signal.

adamgell (commented):
## Exact-SHA independent adversarial review — BLOCK

Reviewed exact base `a8a367c661cd17bd1e73b87724c143ec8565ffc7` through exact head `6629463876c0b08286f2f84deacee0130736dbb8` in a fresh detached worktree. Local HEAD, the remote branch, and the PR head matched before this verdict. I made no repository edits and did not push, resolve threads, approve, or merge.

### Blocking contract gaps

1. **P1 — exact execution keys are prefix matches, not exact token matches.** `crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs:1199-1243` builds `field=value` strings and checks each with raw `str::contains`. An external harness copied the exact `completed` scenario, changed the physical record from `executionId=72400000-0000-0000-0000-000000000005` to `executionId=72400000-0000-0000-0000-000000000005X`, updated only the manifest and expected byte counts, and invoked the real private `validate_contract`; it returned `Ok(())`. A different identifier can therefore satisfy the declared exact key. Parse bounded synthetic fields or require complete token boundaries, and add a permanent fail-closed mutation.

2. **P1 — one finding can mix evidence from both same-time unrelated executions.** At `crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs:1673-1694`, every `insufficientEvidence` finding is considered transaction-bound unconditionally, while other classes require only one overlapping reference. The external harness appended run B evidence to the run A finding in `unrelated-runs`; `validate_contract` returned `Ok(())`. Bind each transaction-scoped finding to exactly one transaction/key and require every cited reference to belong to that transaction; retain a separate source-local path for unkeyed observations.

3. **P2 — the declared correlation boundary is unbound and contradicts the specification.** The specification says `component` is forbidden, and the validator rejects it inside a key at lines 1173-1179, but all 17 checked-in `correlationBoundary.forbiddenJoinFields` arrays omit it. The validator never binds `scope`, `joinFields`, or `forbiddenJoinFields`. The external harness changed `completed` to `scope: crossSideHighConfidence`, `joinFields: [timestamp]`, and an empty forbidden list; `validate_contract` returned `Ok(())`. Bind the declared boundary to the exact enforced constants and add mutation coverage. This also reproduces unresolved exact-head CodeRabbit thread https://github.com/adamgell/cmtraceopen/pull/352#discussion_r3691331414.

4. **P2 — the absent artifact claims an incomplete physical fragment.** `crates/cmtraceopen-parser/tests/fixtures/sccm/client/task_sequence/incomplete/manifest.json:8` combines `captureState: absent` and `relativePath: null` with `rotation.fragmentComplete: false`. No physical fragment exists whose CCM completeness can be false, and `artifact_effective_state` ignores that property for noncaptured states. Omit physical completeness for absent artifacts and make the validator reject it. This is unresolved exact-head CodeRabbit thread https://github.com/adamgell/cmtraceopen/pull/352#discussion_r3691331390.

### Verification evidence

- The correction TDD history is genuine: RED `43cb1f97` ran 10 passed / 3 failed, RED `6a3b8c0e` ran 13 passed / 3 failed, and RED `7c567a4e` ran 16 passed / 2 failed. Exact head runs 18/18 focused green.
- External exact-file adversarial harness: 3/3 reproductions passed, meaning all three fail-open mutations above were accepted.
- Full parser suite: 747/747 passed.
- Rust 1.88 `wasm32-unknown-unknown` check passed.
- Issue-scoped strict Clippy, TypeScript, 34 JSON files, 21 tracked evidence files, LF/privacy/overclaim scans, owned Rust formatting, exact-range `git diff --check`, and clean detached status passed.
- The prescribed strict Clippy command remains red on the same 59 inherited `clippy::uninlined_format_args` findings present at the base. Workspace formatting is likewise red only outside this PR; the owned Rust test is clean.
- The direct-base diff is preparation-only: fixtures, one Rust contract test, and one specification. It changes no parser production source, dependency/build manifest, native collector, database, network, Windows I/O, or parser kind.
- Hosted CodeRabbit posted a substantive exact-head no-findings comment, but two exact-head review threads remain unresolved. Local CodeRabbit retries were rate-limited and are not approval.

This BLOCK is anchored only to `6629463876c0b08286f2f84deacee0130736dbb8`. After TDD corrections, the new exact head still needs substantive CodeRabbit review and a different independent exact-head review. PR #352 must remain draft/open; #319 integration, production reducer work, native discovery/capture, and live Windows acceptance remain separate pending gates.

adamgell:
@coderabbitai full review

crates/cmtraceopen-parser/tests/fixtures/sccm/client/task_sequence/incomplete/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "sccmManifestVersion": 1,
+  "proposalOnly": true,
+  "syntheticFixture": true,
+  "scenario": "incomplete",
+  "bundle": {"role":"client","workflow":"taskSequence","siteCode":"LAB"},
+  "artifacts": [
+    {"artifactId":"task-sequence-incomplete-smsts-absent","designOnlyCatalog":{"entryId":"client-task-sequence-smsts","groupMemberships":["client-task-sequence-smsts"]},"role":"client","kind":"ccmLog","captureState":"absent","encoding":null,"collectionLimit":null,"originalBasename":"smsts.log","sanitizedSourcePath":null,"smstsLogPathEvidence":null,"pathFingerprint":"synthetic:incomplete:candidate","pathClass":"unknown","rotation":{"kind":"current","fragmentComplete":false},"relocationOrdinal":0,"sourceVersion":"5.00.TEST.0000","capturedUtc":"2026-07-30T01:35:00Z","bytesCopied":0,"relativePath":null}
```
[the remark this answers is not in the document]
adamgell: Addressed at head 1db3f4cd via RED d7da8f9a / GREEN 1db3f4cd: incomplete/manifest.json now declares rotation {kind: current} with no fragmentComplete, and the validator rejects any noncapture artifact carrying rotation.fragmentComplete, with a permanent injection mutation test. Focused 22/22, parser 751/751, local CodeRabbit exact-range review of 66294638..1db3f4cd returned no findings. Resolving.

crates/cmtraceopen-parser/tests/sccm_client_task_sequence_fixture_contract.rs
```diff
@@ -0,0 +1,2611 @@
...
+                ));
+            }
+        }
+        for forbidden in ["filename", "path", "timestamp", "displayName", "component"] {
+            if key.contains_key(forbidden) {
+                return Err(format!(
+                    "{transaction_id}: forbidden join field {forbidden}"
+                ));
+            }
+        }
```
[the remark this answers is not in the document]
adamgell: Addressed at head 1db3f4cd via RED d7da8f9a / GREEN 1db3f4cd: the five forbidden names and four exact-key fields are named constants (FORBIDDEN_JOIN_FIELDS, EXACT_KEY_JOIN_FIELDS) reused by the enforcement loops, and the validator binds the declaration: sorted forbiddenJoinFields must equal the enforced five including component (all 17 expected files now declare component), keyed scenarios must declare an allowlisted scope with exactly the exact-key joinFields, unkeyed scenarios stay sourceLocalOnly/coverageOnly. Mutation coverage drives all three drift directions. Resolving.

adamgell (commented):
Independent adversarial review of PR #352 at exact head 1db3f4cd3bbdd26ba9b20956243dc63e0810dc67.
Foundation: merge-base with a8a367c6 is a8a367c661cd17bd1e73b87724c143ec8565ffc7.

Battery: sccm_client_task_sequence_fixture_contract 22 passed; sccm_spine_contract 136 passed;
full cargo test --locked 751 passed, 0 failed; cargo clippy --all-targets -D warnings clean;
wasm32-unknown-unknown check clean. All probes below were run against validate_contract and
reverted; the worktree ends clean at the exact head.

VERDICT: BLOCK. Two of the four previously blocked classes remain partially open.

Gap 1 (exact-token keys, residual admission through bracket delimiters).
complete_field_tokens splits on whitespace and on [ ] < >. Inside a CCM <![LOG[...]LOG]!> body a
bare ] or < is a legal value character (only the full ]LOG]!> sequence terminates the body), so
these characters cannot safely act as token boundaries. Executable evidence: mutating the completed
evidence to runContext=osd]stray while expected.json declares runContext=osd validates Ok; likewise
advertisementId=LAB20305<junk> with declared advertisementId=LAB20305 validates Ok. A declared key
that is a strict prefix of the recorded value is admitted whenever the next on-disk character is one
of the four bracket delimiters. The new GREEN test only probes a non-delimiter suffix (005X).
Suggested fix: split candidate values on whitespace only, or require the needle occurrence to be
bounded by whitespace or record edges.

Gap 2 (one-run findings, source-local laundering).
The insufficientEvidence arm accepts any finding whose citations all match sourceLocalObservations,
but nothing prevents declaring observations that alias keyed transaction evidence. Executable
evidence: in unrelated-runs, adding two keyConfidence \"candidate\" observations citing the run A and
run B records (the same records cited as exact-keyed transaction evidence) and pooling both refs
into one insufficientEvidence finding validates Ok, reopening exactly the cross-run mixing this head
claims to close. Suggested fix: reject any sourceLocalObservation whose evidence ref equals any
transaction evidence, ordering, or terminal ref (observations exist for records that could not be
keyed), then the laundering path disappears.

Closed and verified at this head: boundary binding (8/8 adversarial variants fail closed, including
missing component, extra field, subset, reordered, unkeyed non-empty joinFields, and scope
escalation) and absent rotation (fragmentComplete, encoding, and collectionLimit all rejected on
noncapture artifacts; no truncated field exists in this schema family; the lane is stricter than the
deployment and health siblings). The lastSuccessfulPhase derivation gap from the earlier round was
closed at c8e597e2 and is present at head with mutation coverage. Privacy sweep of the lane fixture
directories and the 66294638..HEAD diff found nothing.

Join-fields design question: requiring declared joinFields to equal the enforced exact-key set is
correct for this corpus. The per-transaction validator unconditionally requires all four fields
(executionId, taskSequencePackageId, advertisementId, runContext) on every keyed transaction, so
equality is the only truthful declaration; unkeyed scenarios take the other branch and must declare
an empty list. No legitimate keyed scenario joins on a subset here, and binding both sides to the
same named constant forces declaration and enforcement to change together. The earlier local
CodeRabbit Major on this check was a false positive. Minor observation: the joinFields comparison is
order-sensitive while forbiddenJoinFields is sorted before comparison; this is inconsistent but only
in the fail-closed direction.

Additional observations (non-blocking): the shipped absent artifact in incomplete still carries
capturedUtc and sourceVersion, which the validator admits (parity with siblings); the manifest and
expected schemas are open, so unknown keys such as truncated or sameTimestampDoesNotJoin are
silently accepted.

This review applies to exact head 1db3f4cd3bbdd26ba9b20956243dc63e0810dc67; any new commits require
re-review.

adamgell:
@coderabbitai full review

adamgell:
@coderabbitai full review

adamgell (commented):
Round 9 review at exact head 495a36ae6e5b5df7dabbaa563b875b526e78b1fc.

Verdict: BLOCK. Two numbered gaps, one of which voids stated claim (2).

Descent
git merge-base HEAD a8a367c661cd17bd1e73b87724c143ec8565ffc7 prints
a8a367c661cd17bd1e73b87724c143ec8565ffc7, so the foundation is an
ancestor of head.

Battery, all green
- sccm_client_task_sequence_fixture_contract: 24 passed, 0 failed
- sccm_spine_contract: 136 passed, 0 failed
- cmtraceopen-parser full suite: 753 passed, 0 failed
  (cargo test --workspace: 1507 passed, 0 failed)
- cargo clippy --workspace --all-targets -- -D warnings: exit 0
- cargo check and cargo clippy for wasm32-unknown-unknown: exit 0
- Privacy sweep of 66294638..HEAD: clean. The JSON diff is only the
  component addition to 17 forbiddenJoinFields lists plus one
  fragmentComplete removal on the noncapture artifact.

Claim (3), sort-insensitive joinFields: CLOSED.
Reordered declaration accepted. Wrong set (component substituted),
duplicate entry, superset, and subset all rejected with \"declared join
fields do not match the enforced exact key fields\".

Claim (1), token boundaries: closed for every in-body case, but the
guarantee is narrower than stated. See gap 2.
Accepted: needle bounded by the genuine terminator after _SMSTSLogPath
is moved ahead of it; needle at the exact start of the body with no
leading whitespace; tab separator instead of space; CRLF line endings.
Rejected with \"co-occur\": a value containing ]LOG]! without the closing
angle bracket; two records on one physical line with the key only in the
second; and both round-7 probes (runContext=osd]stray,
advertisementId=LAB20305<junk>).

Claim (2), observation disjointness: NOT achieved. See gap 1.

Gap 1 (blocking). Padded evidence references defeat the new
disjointness rule and re-open the round-7 one-run closure.

The check at sccm_client_task_sequence_fixture_contract.rs:1678-1694
compares whole serde_json Value objects
(transaction_ref == &observation[\"evidence\"]). Evidence references are
open objects and evidence_text reads only artifactId, startLine, and
endLine, so adding one ignored key evades the comparison while the
citation still resolves to the same physical record.

Probes on the unrelated-runs scenario:
- control, unpadded alias of keyed run A evidence: REJECTED with
  \"keyed transaction evidence cannot double as a source-local
  observation\"
- same alias with one extra ignored key added to the evidence object:
  ACCEPTED
- one finding citing evidence from BOTH unrelated exact runs, laundered
  through two padded source-local observations: ACCEPTED

The third probe is exactly the shape that
finding_evidence_cannot_mix_unrelated_exact_runs (line 3035) pins, and
it passes at head. The same open-object equality is load-bearing in
cited_refs_are_source_local, binding_transactions, and the ordering and
terminal membership checks, so the entire binding chain shares this
weakness.

Suggested fix: compare the identity triple
(artifactId, startLine, endLine), or reject evidence-reference objects
that carry keys outside that triple.

Gap 2 (blocking as claimed, but pre-existing rather than a regression).
The complete_field_tokens fallback widens the token universe to the
whole raw line.

complete_field_tokens ends in .unwrap_or(record_text). When the cited
record text does not start exactly with the CCM opener, the function
silently tokenizes the entire raw line, including the
<time=... context=\"...\" ...> trailer.

Probes on the completed scenario:
- a single leading space with the record otherwise untouched: ACCEPTED,
  so the fallback is reachable and the framing and fragmentComplete
  rules do not catch it
- all four exact key values removed from the body and relocated into the
  quoted context trailer attribute, with a leading space: ACCEPTED
- the identical mutation without the leading space: REJECTED with
  \"co-occur\"

This is not a regression. Reimplementing the pre-495a36ae tokenizer
finds all four needles in the same exploit string, while the round-7
osd]stray probe is false at head and true pre-495a36ae, confirming the
round-7 improvement is real. The residual is that claim (1) as written
promises body bounding unconditionally, and the fallback silently
degrades it to whole-line matching.

Suggested fix: fail closed. Return an empty token set, or surface an
explicit error, when the record body cannot be delimited. The framing
check already guarantees the cited range is one complete CCM record, so
an undelimitable body is a contract violation rather than a reason to
widen.

Non-blocking observations
- The orderingEvidence and terminalEvidence disjuncts in the new
  disjointness check are unreachable. Both are already required to be
  members of the evidence array; a probe that moves orderingEvidence out
  of the array is rejected by the membership rule first. Harmless
  redundancy, worth a comment.
- Two observations with distinct IDs citing an identical record are
  admitted. No authority gain was found.
- A declared value whose raw text contains an embedded terminator
  (runContext=osd]LOG]!>x) is accepted. This is consistent with the
  parser's own framing and is the same rule that makes the required
  genuine-terminator nuance pass. Not a defect.
- A second CCM opener appearing mid-body is accepted. Defensible: the
  needle is genuinely whitespace-bounded inside the framed body.
- The open-schema posture noted in round 7 is normally program-wide and
  not a lane blocker, but gap 1 shows it converting into a bypass of two
  closures this series explicitly claims, which is what makes it
  blocking here rather than cosmetic.

Rounds 1-8 closures, independently reconfirmed with fresh mutations
rather than the shipped assertions alone
- exact-token prefix and suffix (runContext=xosd, runContext=osdx):
  both rejected
- absent rotation on the noncapture artifact (fragmentComplete,
  encoding, collectionLimit, bytesCopied): all four rejected
- boundary binding with component removed from forbiddenJoinFields:
  rejected
- lastSuccessfulPhase ahead of the observed phase, and a bogus phase:
  both rejected
- observation off-by-one is not over-broad: an observation citing a
  neighbouring unkeyed record is accepted, while an observation citing
  the keyed record on line 1 is rejected in the same scenario

Exact-head statement
Every command above was run in a detached worktree pinned to
495a36ae6e5b5df7dabbaa563b875b526e78b1fc. All probes were temporary and
were reverted; git status --porcelain is empty and the contract file is
byte-identical to the committed version at that head.

Controller note: this review was performed at 495a36ae. Head has since advanced to ccda4dfd (a rotation capture-provenance RED/GREEN pair). Both gap sites were verified unchanged at ccda4dfd: the fallback is still at line 425 (`.unwrap_or(record_text)`) and the whole-Value comparison is still at line 1703 (`transaction_ref == &observation["evidence"]`). Both findings therefore apply to the current head as written.

adamgell:
@coderabbitai full review

adamgell (commented):
Round 10 review at exact head d1d163b3069b5c00c8d9b17663ecac884fd8efbe.
Reviewed in a detached worktree pinned to that commit; the tree was clean at
the start and at the end of the review. Foundation SHA from
git merge-base HEAD a8a367c661cd17bd1e73b87724c143ec8565ffc7 is
a8a367c661cd17bd1e73b87724c143ec8565ffc7. The whole contract test file is
added by this PR, so every line discussed below is PR-owned.

Battery, exact counts, all green at d1d163b3:
  cargo test -p cmtraceopen-parser --test sccm_client_task_sequence_fixture_contract
    29 passed, 0 failed
  cargo test -p cmtraceopen-parser --test sccm_spine_contract
    136 passed, 0 failed
  cargo test -p cmtraceopen-parser
    758 passed, 0 failed
  cargo test --workspace
    green
  cargo clippy --workspace --all-targets -- -D warnings
    no warnings, no errors
  cargo check -p cmtraceopen-parser --target wasm32-unknown-unknown
    clean
  rustfmt --check --edition 2021 on the single changed file
    exit 0

RED discipline. Restoring the source state of 531e8e85 fails exactly the three
new tests and no others (26 passed, 3 failed):
  padded_evidence_references_cannot_alias_keyed_records
  padded_observations_cannot_launder_cross_run_findings
  undelimitable_record_bodies_fail_closed
Each failure is an expect_err on an Ok(()), which shows the pre-fix validator
accepted the padded alias, the cross-run laundering, and the relocated-key
shape rather than merely reporting a different error. Restored to d1d163b3,
29 of 29 pass.

Mechanism 1, validate_evidence_reference_shapes. Verified closed.
Padded references were probed at every site an evidence reference can occupy
in a shipped expected document, plus two synthetic containers:
  transactions[].evidence[]                          rejected: unmodeled fields
  transactions[].orderingEvidence                    rejected: unmodeled fields
  transactions[].terminalEvidence                    rejected: unmodeled fields
  sourceLocalObservations[].evidence                 rejected: unmodeled fields
  findings[].evidence[]                              rejected: unmodeled fields
  logicalReconstructions[].smstsLogPathEvidence      rejected: unmodeled fields
  novel key, depth 6, through arrays of arrays       rejected: unmodeled fields
  hung off correlationBoundary                       rejected: unmodeled fields
Coverage is total by construction because the sweep recurses every object value
and every array element from the expected root.

Evasion by omission or by type does not open a hole. Nine variants were probed
(missing endLine, missing startLine, missing artifactId, startLine as a string,
startLine as a float, negative endLine, numeric artifactId, identity nested one
level deeper, and a pure key reordering). The first eight all fail closed,
because evidence_text, manifest_artifact and collect_evidence_refs read through
the same as_str and as_u64 accessors as evidence_reference_identity, so a
reference that escapes the identity extractor also cannot resolve to a physical
record. The reordering variant is accepted, which is correct: it is a
legitimate identity.

Sweep ordering is correct. validate_evidence_reference_shapes is the second
statement in validate_contract and the only earlier call, validate_manifest_and
_storage, reads the manifest and the filesystem and never touches expected. A
document with workflow, contractState and extractionProfile.status all broken
plus one padded terminalEvidence still fails with the unmodeled-fields error,
so no earlier check can be fooled first.

The characterization of same_evidence_reference as defense in depth rather than
the active catch is accurate, though the stated reason is not quite right on
this build. cargo tree shows serde_json/preserve_order is enabled under a
workspace-wide build through the tauri graph and is not enabled under
-p cmtraceopen-parser, so the map type differs between the two invocations.
Both IndexMap and BTreeMap compare JSON objects order-insensitively, so whole
value equality was never order-sensitive either way. Verified executably: whole
value equality and same_evidence_reference both return true for a reordered
identity.

Mechanism 2, complete_field_tokens returning Option. Verified closed for key
admission. Eight framing shapes were probed against the completed scenario:
  opener present, terminator removed        rejected by the CCM grammar check
  terminator present, opener corrupted      rejected by the CCM grammar check
  terminator placed before the opener       rejected by the CCM grammar check
  leading tab before the opener             rejected: not delimited by the CCM framing
  leading CR before the opener              rejected: not delimited by the CCM framing
  empty body between valid delimiters       rejected by the _SMSTSLogPath check
  second opener/terminator pair with the key fields moved after the first
    terminator                              rejected: key fields do not co-occur
  CRLF line endings, record otherwise untouched   accepted, which is correct
The two shapes that reach the new None branch are precisely the round 9 family.
The embedded second pair confirms that the first terminator wins, so an
injected pair cannot smuggle key tokens.

The round 7 genuine-terminator case has not regressed. After removing
_SMSTSLogPath from the WinPE record the body ends runContext=osd]LOG]!>; the
scenario still validates, complete_field_tokens still yields runContext=osd,
and no trailer field such as component= appears in the token set.

Rounds 1 through 9 were re-confirmed with ten fresh mutations, all rejected:
strict-prefix advertisementId, suffixed taskSequencePackageId, strict-suffix
runContext, lastSuccessfulPhase advanced past the observed phase, narrowed
forbiddenJoinFields, narrowed joinFields, a single finding pooling both exact
runs with unpadded references, rotated-fragment provenance renamed to the
current file, the current fragment declared as the rotated one, and a
reconstruction ordered against the rotation order.

Privacy. git diff over 66294638..HEAD and over the full PR range shows no
hostnames, user paths, drive-letter paths, IP addresses, mail addresses, SIDs,
credentials or tenant identifiers in added lines. Every fixture path is a
SYNTHETIC:// URI. The only pattern hits in the range are inside
fixture_privacy_and_scope_boundaries_are_pinned, which asserts those patterns
are absent from the corpus.

Blocking gap 1. The record trailer that the framing fix removed from key
admission is still admitted for phase, state and terminal binding.

complete_field_tokens now yields tokens only from the delimited body, but the
same loop pushes the raw record_text into cited_record_texts, and the phase and
state binding at lines 1402 to 1404 and the terminal binding at lines 1699 to
1701 use a raw substring contains over the whole physical line, including the
<time=... context=\"...\" ...> trailer. Three mutations of the completed scenario,
each with only bytesCopied resynced in the manifest and the expected document,
were accepted while the contract continued to declare the strongest claim in
the corpus (classification success, confidence high, state succeeded, phase
complete, terminal outcome):

  1. phase=complete state=succeeded deleted from the record body and placed in
     context=\"phase=complete state=succeeded\". Accepted. The body then states
     nothing about phase or state.
  2. Body rewritten to phase=completeX state=succeededX with the contract left
     declaring complete and succeeded. Accepted, because contains has no token
     boundary.
  3. terminal=true deleted from the body and placed in context=\"terminal=true\".
     Accepted. Separately, a body carrying state=succeededLater satisfies the
     terminal state=succeeded check.

This is the same laundering class the last several rounds have been closing,
in the same loop, and it defeats this PR's own assertion in
adversarial_contract_mutations_fail_closed that phase must bind to cited CCM
evidence. The doc comment added to complete_field_tokens names this trailer as
the thing that must not widen admission.

Suggested fix, mechanical: bind phase, state and terminal from
complete_field_tokens applied to the delimited body of the cited record and of
the terminal record, instead of contains over the raw line, so the same body
edges and the same complete-token rule govern all three. Checked across every
shipped fixture record: phase=, state= and terminal= occur only inside the body
and only as complete whitespace-delimited tokens, so this change should require
no fixture edits.

Non-blocking observation. The identity-triple convention is enforced only
inside this test binary. crates/cmtraceopen-parser/tests/fixtures/sccm/client/
policy/contradictory-offset/expected.json ships triple-bearing objects that
carry additional keys at offsetOrderingContract.orderedEvidence[0],
offsetOrderingContract.orderedEvidence[1] and
offsetOrderingContract.nonComparableEvidence[0]. Those belong to a different
corpus with its own validator and are unaffected by this PR, but the convention
is worth stating explicitly if it is meant to become general.

Verdict: BLOCK at d1d163b3069b5c00c8d9b17663ecac884fd8efbe. Both round 9 gaps
are closed and verified against the new mechanisms directly rather than by
replaying round 9's probes. One numbered gap remains, reproducible from the
shipped completed fixture with three independent mutations.

adamgell:
@coderabbitai full review

adamgell (commented):
Round 11 review at exact head 63e4ad41 (RED c0486b3f / GREEN 63e4ad41).

BATTERY
focused 30 passed / 0 failed
spine 136 passed / 0 failed
parser crate 759 passed / 0 failed (351+222+8+3+3+30+1+5+136)
workspace 1513 passed / 0 failed
cargo clippy --workspace --all-targets -- -D warnings: exit 0
cargo check -p cmtraceopen-parser --target wasm32-unknown-unknown: clean
git diff --check 66294638 HEAD: clean
rustfmt --check on the changed file: clean

RED DISCIPLINE
Restoring the c0486b3f source state reproduces 29 passed / 1 failed with the
single table-driven test naming all six shapes: phase-and-state-relocated-to-
trailer, terminal-relocated-to-trailer, phase-token-suffixed, state-token-
suffixed, terminal-token-suffixed, terminal-state-token-extended. Restored to
clean afterward. The 30 rather than 32 count is correct: the six shapes belong
in one table following the existing coherent_review_mutations_fail_closed idiom.

NOVEL SHAPES
Twelve shapes beyond the shipped table, all fail closed at head:
leading-character variants on token and value for phase, state and terminal;
body-wrong/trailer-right for phase, state and terminal; phase and state split
across cited records so no single record carries both; the whole claim
relocated onto a different cited record; terminal record with a delimitable
body but state= or phase= only in the trailer. Body-right/trailer-wrong is
correctly accepted, which settles the precedence question: the body is
authoritative and the trailer is ignored rather than merged.

INDEPENDENT AUDIT
Every consumer of record text and every contains over a record line was
enumerated. The two routed sites are correct. The four exclusions are sound.

The load-bearing exclusion was verified directly rather than accepted. od -c on
the rotation-boundary lo fragment shows it opens <![LOG[, carries
_SMSTSLogPath=, and contains zero ]LOG]!> terminators. Body-bounding the
logical path citation site breaks rotation-boundary; body-bounding the
whole-file scan site breaks rotation-boundary and leading-space. Both
exclusions are necessary, and both are fail-closed in direction: an extra
_SMSTSLogPath anywhere in the bytes breaks the exact-count requirement rather
than laundering a claim.

The remaining three exclusions hold. The two evidence_text calls for
observation and finding citations discard their result and perform no text
matching. The notasksequence check scans finding object keys, not record text.
The privacy scan must remain a raw whole-file scan or it stops seeing the
trailer. No missed site was found; the post-fix grep for record_text.contains,
terminal_text.contains and _text.contains returns none.

OBSERVATION (not blocking, suggest a follow-up issue)
complete_field_tokens tokenizes with split_whitespace(), which treats tab,
NBSP (U+00A0) and form feed as separators, so a value extended past one of
those characters still satisfies a shorter declared value. This is not
introduced here: the helper predates this change and already backs key
admission, the pre-fix raw contains predicate accepted every one of these
cases as well, so post-fix admission is a strict subset of pre-fix admission,
and planting such a character in a shipped fixture trips the pinned corpus
digest (evidence_bytes 8249 against the pinned 8243, plus a digest mismatch).
Worth tightening in the helper separately rather than gating this round.

PRIOR ROUNDS
Ten fresh mutations against the rounds 1 through 10 closures all fail closed:
extended key value, prefixed key token, key relocated to trailer,
bracket-bounded key prefix, removed body terminator, terminal evidence not
key-bound, widened evidence range, swapped rotation basename, inadmissible
lastSuccessfulPhase, dropped timestamp precision.

PRIVACY
66294638..HEAD changes no evidence bytes. All added lines are ASCII. Only the
synthetic 72400000-0000-0000-0000-000000000005 GUID appears. No host paths,
emails, domains, credentials or secrets.

VERDICT: PASS at exact head 63e4ad41. Both round-10 gaps are closed, the class
audit is accurate, and the worktree was verified clean at start and end.

Issue #376, opened 2026-07-31
Application-wide restart as administrator with source restoration


adamgell:
# Application-Wide Restart as Administrator

## Status

Approved by the repository owner on 2026-07-30.

The approved trigger model is both:

1. an explicit application-wide menu action; and
2. a contextual offer after a source operation fails with Access Denied.

The approved restoration model is limited to the active workspace/source that
requested elevation. CMTrace Open must not restore every open tab or serialize
the entire application session.

## Outcome

CMTrace Open will provide one safe, application-owned Windows elevation
mechanism that can be invoked from anywhere in the application.

Users can:

- choose **Restart as administrator** from the global application menu; or
- choose the same action from an Access Denied recovery prompt.

After Windows approves the UAC request, the non-elevated process exits and the
elevated process restores only:

- the active workspace; and
- the active file, folder, or known-source intent when one exists.

The existing ESP Diagnostics elevation banner will reuse this mechanism rather
than own a separate relaunch implementation.

## Problem

The current implementation is ESP-specific:

```text
src-tauri/src/esp/relaunch.rs
restart_esp_as_administrator
restartEspAsAdministrator
ElevationBanner
--workspace=esp-diagnostics
```

The native ShellExecute `runas` implementation is already security-conscious,
but it always routes to ESP Diagnostics and deliberately discards all file
paths. That prevents other workspaces and protected log sources from using the
same capability.

CMTrace Open now opens protected Windows sources outside ESP, including logs
under Program Files, Windows system directories, registry-backed workflows,
and future SCCM server/client workspaces. A user should not need to know in
advance which source requires elevation or manually reconstruct their work
after restarting the application.

## Scope

This change includes:

1. a generic application elevation module and IPC command;
2. a global Windows menu action;
3. a contextual Access Denied recovery prompt;
4. one-time restoration of the requesting workspace/source;
5. migration of the ESP banner to the generic command;
6. stable structured access-denied classification;
7. cancellation, failure, stale-state, and loop prevention;
8. focused Rust, frontend, startup-routing, and Windows-native tests.

This change does not include:

- starting CMTrace Open elevated by default;
- silently displaying a UAC prompt without a user click;
- restoring all open tabs, filters, selected rows, searches, or UI layout;
- forwarding authentication tokens, Graph state, evidence contents,
  credentials, or arbitrary command-line arguments;
- changing Windows ACLs;
- bypassing UAC;
- auto-retrying privileged writes that the user did not explicitly request;
[...]

PR #382, opened 2026-07-31
feat(elevation): application-wide restart as administrator

adamgell:
Implements the backend, frontend, and menu surface of #376.

## What this does

Generalizes the ESP-specific relaunch into one application-owned elevation
mechanism any workspace can use. Before this, `src-tauri/src/esp/relaunch.rs`
owned the only `ShellExecute`/`runas` implementation, always routed to
`esp-diagnostics`, and deliberately discarded every file path — so a user who
hit a protected source anywhere else had to know to close the app and relaunch
it themselves.

### Backend

- `elevation/mod.rs` — the validated request contract. `AppWorkspace` mirrors the
  frontend `WorkspaceId` union as an allowlist; a `RestoreTarget` carries at most
  one bounded, absolute, control-character-free path or one conservative catalog id.
- `elevation/relaunch.rs` — platform mechanics, with no dependency on any product
  module so it survives `esp-diagnostics` being compiled out. Child arguments are
  rebuilt from nothing: fixed `runas` verb, current executable, and at most one
  validated `--elevation-restore=<uuid>`.
- `elevation/restore_ticket.rs` — the handoff. The child gets only an opaque id;
  the typed intent lives in a schema-versioned, expiring, size-bounded file under
  an app-owned per-user directory. Claimed by rename so it is consumed exactly
  once, refuses symlinks and non-regular files, and is re-validated on read
  because per-user storage is still untrusted input.
- The Windows token elevation probe moves out of the ESP system provider so there
  is one owner.

### Frontend

- `src/lib/elevation.ts` is the single coordinator. It never throws — every
  failure, cancellation, and unsupported platform is an outcome — and concurrent
  requests collapse, so UAC prompts cannot stack.
- Startup routing applies the documented precedence: positional file paths, then
  a valid restore ticket, then the legacy workspace argument, then default.
- A ticket restores one workspace and at most one source. Other tabs, filters,
  searches, and selected rows are not restored.
- `markElevationRetryAttempted` fires before restoration, so a restored source
  that is still denied offers troubleshooting rather than a second prompt.

### Menu

`File > Restart as Administrator…` on Windows only, in its own group above Exit.
Absent on macOS and Linux rather than shown disabled: muda has no `set_visible`,
so an action that can never apply must not be appended at all. Confirmation is
requested before the backend is touched, so UAC cannot appear without a click.

## Verification

```
cargo test --locked --manifest-path src-tauri/Cargo.toml --lib   481 passed
cargo clippy --locked --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings   clean
npx tsc --noEmit    clean
npm test -- --run   553 passed
```

## Deliberately not done

- **Structured Access Denied classification and the recovery prompt.** The issue
  requires the frontend not to decide on elevation by matching localized strings,
  which needs a typed classification on the shared log-source loading path.
  `AppError` currently serializes to a plain string via `InvokeError::from`, so
  this needs its own change rather than being bolted on here.
- **Documentation** (issue phase 7).
- **Native Windows acceptance pass.** Cannot be run from this macOS host; the
  Windows-gated code is covered by `#[cfg(target_os = "windows")]` unit tests that
  only execute in CI's Windows job.

## Assumptions

- `cargo fmt --check` is not a CI gate here and the tree has pre-existing drift, so
  only files this PR creates were formatted. Unrelated files are untouched on purpose.
- ESP's `restart_esp_as_administrator` command is left in place for now; the banner
  already routes through the generic coordinator, so there is one backend behavior
  owner even though the old command still exists.

Refs #376

### DIFF
--- a/src-tauri/src/commands/elevation.rs
+++ b/src-tauri/src/commands/elevation.rs
@@ -1,0 +1,279 @@
+//! Tauri IPC for application-wide restart as administrator.
+//!
+//! This is the only place the application decides to relaunch itself elevated.
+//! Every workspace, the global menu, and the Access Denied recovery prompt go
+//! through `restart_as_administrator`; none of them owns a relaunch of its own.
+
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use serde::Serialize;
+use tauri::{AppHandle, Manager, State};
+use thiserror::Error;
+
+use crate::elevation::relaunch::{
+    restart_with_provider, NativeRelaunchProvider, RelaunchError, RelaunchReason, RelaunchResult,
+};
+use crate::elevation::restore_ticket::{
+    consume_ticket, discard_ticket, prune_expired, ticket_directory, ticket_for, RestoreTicket,
+    TicketError,
+};
+use crate::elevation::{AppElevationState, ElevationRequest, ElevationValidationError};
+use crate::state::app_state::AppState;
+
+/// Guards against a double-click or two workspaces racing the same prompt.
+///
+/// Only one request may be in flight: a second concurrent call is refused
+/// rather than queued, so the user cannot stack UAC prompts.
+static REQUEST_IN_FLIGHT: AtomicBool = AtomicBool::new(false);
+
+/// Releases the in-flight guard however the request ends, including on an early
+/// return or a panic unwinding through the command.
+struct InFlightGuard;
+
+impl InFlightGuard {
+    fn acquire() -> Option<Self> {
+        REQUEST_IN_FLIGHT
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .ok()
+            .map(|_| Self)
+    }
+}
+
+impl Drop for InFlightGuard {
+    fn drop(&mut self) {
+        REQUEST_IN_FLIGHT.store(false, Ordering::SeqCst);
+    }
+}
+
+/// Why a restart request could not be carried out.
+///
+/// Nested causes stay nested rather than flattened: both this enum and its
+/// sources are internally tagged on `kind`, so flattening would collide.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Error)]
+#[serde(tag = "kind", rename_all = "camelCase")]
+pub enum ElevationCommandError {
+    #[error("another administrator restart is already in progress")]
+    AlreadyInProgress,
+    #[error("the elevation request was rejected: {reason}")]
+    InvalidRequest { reason: ElevationValidationError },
+    #[error("the elevation restore ticket could not be prepared")]
+    TicketUnavailable,
+    #[error("{source}")]
+    Relaunch { source: RelaunchError },
+    #[error("the application state directory is unavailable")]
+    StateDirectoryUnavailable,
+}
+
+impl From<ElevationValidationError> for ElevationCommandError {
+    fn from(reason: ElevationValidationError) -> Self {
+        Self::InvalidRequest { reason }
+    }
+}
+
+impl From<RelaunchError> for ElevationCommandError {
+    fn from(source: RelaunchError) -> Self {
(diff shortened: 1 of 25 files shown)


adamgell:
### Code review

Multi-lens review over `bc2cd1f..eaa5bf9` (5 independent lenses, each finding independently confidence-scored; 23 raw findings, 2 survived at >=80).

Found 2 issues, both fixed in [`b492754`](https://github.com/adamgell/cmtraceopen/commit/b492754c22e243a827158ab137d3d289ef3b32b3):

1. **`File > Restart as Administrator` was inert** (95, found independently by 3 lenses). The id was added to `FILE_ORDER_WINDOWS` and appended in `build_file_menu`, but never registered in `payload_for_menu_id`, whose match ends in `_ => return None`. Clicking it logged `unrecognized menu_id` and emitted no `app-menu-action`, so the frontend handler was unreachable. Compiled clean, passed clippy.

https://github.com/adamgell/cmtraceopen/blob/eaa5bf989e3faeda565be62d8aa377be243613ce/src-tauri/src/menu.rs#L1726-L1731

   The hand-maintained case list in `every_static_action_has_the_expected_payload` could not catch this, since a list cannot prove an item is missing from it. The fix adds `every_ordered_menu_item_dispatches_an_action`, which walks the ORDER constants and asserts each rendered item resolves to a payload.

2. **`RestoreTarget` crossed the IPC boundary with snake_case fields** (92, found independently by 3 lenses). `rename_all` on an enum renames variants only; struct-variant fields need `rename_all_fields`. `KnownSource { source_id }` serialized as `source_id` while `src/types/elevation.ts` declares `sourceId`, so every known-source elevation request failed deserialization before the handler ran, and the restore-ticket read-back was broken symmetrically.

https://github.com/adamgell/cmtraceopen/blob/eaa5bf989e3faeda565be62d8aa377be243613ce/src-tauri/src/elevation/mod.rs#L82-L87

   Neither clippy nor tsc can see across that boundary, and every existing test constructed the variant in Rust rather than exercising the wire format. Both directions are now asserted against the literal JSON the frontend sends, including that the snake_case form is rejected rather than tolerated.

The known-source catalog is largely protected locations like `C:\Windows\CCM\Logs` — exactly the sources that need elevation — so issue 2 was not a rare path.

Verification after fixes: 484 Rust tests, 553 frontend tests, clippy clean, tsc clean.

adamgell:
### Status and overlap

Phase 6 (structured Access Denied classification and recovery prompt) landed in `8ab40fe`. Phases 1-6 of the issue are now implemented here; phase 7 (documentation) and the native Windows acceptance pass are not.

**This PR overlaps #384**, which was opened during the same window and implements all 7 phases including documentation. Both close #376. They converged independently on the same module paths, the same three IPC command names, the same `--elevation-restore=<uuid>` flag, and the same ticket layout — which is reassuring about the design but means one of these should be closed rather than both merged.

#384 is the more complete of the two. What is here and not obviously there:

- `every_ordered_menu_item_dispatches_an_action` — walks the menu ORDER constants and asserts every rendered item resolves to a payload. A hand-maintained id list cannot prove an item is missing from it, which is exactly how the inert-menu-item bug got in here in the first place.
- `rename_all_fields = "camelCase"` on `RestoreTarget`, with tests asserting the literal JSON in both directions including that the snake_case form is rejected. Flagged on #384 as [a comment](https://github.com/adamgell/cmtraceopen/pull/384#issuecomment-5139644942) since it appears to apply there too.
- The single-probe restructure in `classifyCommandError`, so adding the access-denied classification does not hand a hostile Proxy a second `getPrototypeOf` trap invocation.

Happy for this to be closed in favour of #384 with those three lifted across.

adamgell:
Superseded by #384, which contains these same four elevation commits plus Access Denied classification and the restore dialog. Closing to leave a single implementation of #376.

Issue #328, opened 2026-07-30
SCCM Server: diagnose management-point client transactions


adamgell:
## Parent

Part of #317.

## Outcome

Corroborate or isolate management-point registration, authentication, location, policy, relay/status, and notification failures.

## Initial input contract

- Management-point artifacts including MP_CliReg, MP_GetAuth, MP_GetPolicy, MP_Location, MP_RegistrationManager, MP_Status, MP_Framework, CcmIsapi, CCM_STS, ClientAuth, and MP control evidence.
- Optional client request/response artifacts with matching stable identifiers.

## Acceptance criteria

- Build request/response transactions for registration, auth, location, policy, and status flows.
- Extract client identity, request/topic, assigned site, and role/host keys while preserving privacy-safe context.
- Distinguish no request received, authentication/authorization failure, location response failure, policy response failure, MP control/role health failure, and incomplete capture.
- Correlate to client evidence only with matching keys; time proximity alone remains low confidence.
- Add server-only and cross-side fixtures with terminal, retry, and partial-coverage outcomes.

## Boundary

This is not a generic IIS parser replacement; existing IIS evidence is supplemental.

Microsoft reference: https://learn.microsoft.com/en-us/intune/configmgr/core/plan-design/hierarchy/log-files

adamgell:
Execution claim — preparation slice active

- Agent worktree: `.worktrees/sccm-328-mp-corpus`
- Branch: `codex/sccm-328-mp-corpus`
- Base: `f2caccfb8602638807bff5c19d722c037676ca43`
- Ownership: #328 preparation document and `fixtures/sccm/server/management-point/**` only.

The active agent is building the plan-prescribed sanitized MP scenario matrix and focused contract proof test-first. Production reducer/native adapter work remains blocked on reviewed #318/#335 contracts. Producer role/topology provenance is retained; missing default paths and time proximity cannot establish role health or client/server causality. A branch/commit/PR update will be posted after the focused and aggregate gates pass.

adamgell:
Prepared #328 and opened ready PR #343 against `codex/parser-family-skeleton`.

## Delivery

- Worktree: `/Users/Adam.Gell/repo/cmtraceopen/.worktrees/sccm-328-mp-corpus`
- Branch: `codex/sccm-328-mp-corpus`
- Commit: `7a40fc8429417f5d2dd3a237e938ecf477bb6956`
- Commit subject: `docs(sccm): prepare management point corpus`
[...]

Issue #327, opened 2026-07-30
SCCM Server: diagnose site-core and status-system workflows


adamgell:
## Parent

Part of #317.

## Outcome

Diagnose site-server component health, inbox processing, status/state messaging, and monitoring without guessing from a single component error.

## Initial input contract

- Site-core records such as SMS Executive, Site Component Manager, Hierarchy Manager, inbox assistant/monitor, component monitor, site status, Status Manager, and State System evidence.
- Optional exported status-system messages and site configuration facts with explicit provenance.

## Acceptance criteria

- Classify site-component startup/configuration, inbox receive/process/backlog, component health, status processing, and state-message processing phases.
- Correlate role/component/site identifiers and status messages deterministically.
- Distinguish a component symptom, terminal component failure, stale/backlogged work, and insufficient coverage.
- Add multifile fixtures for healthy processing, backlog/retry, terminal failure, conflicting evidence, rotations, and partial collection.
- Cite entries and request the exact related role bundle rather than attributing all site errors to a client.

## Boundary

Management-point, DP, SUP, hierarchy transport, and provider/Admin Service paths have separate issues.

Microsoft reference: https://learn.microsoft.com/en-us/intune/configmgr/core/plan-design/hierarchy/log-files

adamgell:
### Agent claim / corpus review state

**Owner:** server site-core corpus agent, branch `codex/sccm-327-site-core-corpus`, commit `49ad6f1`.

**Fixture matrix:** healthy, component failure, inbox backlog, status-processing failure, ordered same-key recovery, contradictory keyed transactions, malformed, rotation boundary, and incomplete — 9 scenarios / 18 artifacts / 14 evidence files.

**Verification:** exact-byte corpus validator passes; 18 JSON files parse; complete fixtures are valid CCM; parser suite 571 passed in owner run; strict Clippy, wasm32, TypeScript, and diff checks pass. No production reducer code is included.

**Review result:** one Important fixture-contract defect remains: the physically incomplete artifact’s `expected.json` marks an incomplete logical record, but its manifest omits `rotation.fragmentComplete: false`. A new isolated fix branch/worktree will own that one correction and a focused re-review.

**Dependency state:** production reducer remains blocked on reviewed #318 and #335 contracts. No native Windows acceptance or cross-side causality claim.

adamgell:
### Review fix published

[...]

PR #405, opened 2026-08-01
feat(sccm): analyze site core status evidence

adamgell:
Implements the server site-core and status-system workflow analyzer for #327, specified by the already-merged site-core corpus.

## Public API

New `sccm/server/windows/site_core.rs`, re-exported through `windows/mod.rs`:

`analyze_site_core(&SccmSiteCoreBundle) -> SccmSiteCoreAnalysis`

Input is a typed bundle (`SccmSiteCoreTopology`, sources carrying artifact plus source group, rotation lineage, fragment completeness, physical line end, and capture limit, plus normalized evidence). Output carries schema version, workflow, profile, a five-phase state chain, results, unlinked observations, coverage gaps, findings, artifact requests, prohibited claims, and a cross-side correlation flag. All serialized fields are additive and camelCase, and `SccmSiteCoreFinding` flattens the shared `SccmFinding` so every finding passes the spine's `validate()` on serialization.

## Corpus coverage

All nine merged scenarios reproduce exactly: healthy, component-failure, inbox-backlog, status-processing-failure, recovery, contradictory, incomplete, rotation-boundary, and malformed. Notable conservatism the corpus pins: component-failure emits no artifact request because the failure is already confirmed; recovery requires a strictly later terminal success under comparable UTC; contradictory keeps two separate results because same-instant records for different components never merge; and both rotation-boundary and malformed produce zero results, surfacing observations plus a request instead.

## Verification

- `--test sccm_server_site_core`: 13 passed
- `--test sccm_site_core_fixture_contract`: 5 passed
- `--test sccm_spine_contract`: 137 passed
- `cargo test --locked -p cmtraceopen-parser`: 938 passed, 0 failed across 18 targets
- `cargo clippy --locked -p cmtraceopen-parser --all-targets -- -D warnings`: clean
- `cargo check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown`: clean, no new dependencies, no I/O, no async, no Windows or Tauri usage
- `git diff --check 26116027 HEAD` clean; `rustfmt --check` clean on all three changed files; `npx tsc --noEmit` exit 0

`intake.rs` and `catalog.rs` are untouched. `mpcontrol.log` cannot be admitted here because `server-mp-policy` is not a site-core group, preserving the boundary lane #328 established.

## Deliberately not implemented, with reasons

**1. The reducer does not consume `assess_server_intake`, because the two schemas are currently incompatible.** This is a real corpus-versus-code disagreement that needs a decision outside this lane, not a shortcut. Intake requires `producerRole`, `sourceId`, `configuredPathProvenance.pathFingerprint`, `rotation.lineageId`, `collectionLimit`, a closed artifact-id allowlist, `originalPath` matching `REDACTED_*`, and `evidence/sccm/server/site-server/server-sitecomp/...` paths. The merged site-core corpus uses `role`, `sourceGroup`, `configuredPath` as a bool, `rotationLineage`, `captureLimitBytes`, ids such as `healthy-sitecomp-current`, `originalPath: "REDACTED"`, and `evidence/sccm/server/site-core/sitecomp/...` paths. Feeding a site-core fixture to `assess_server_intake` fails at several independent gates. This PR follows lane #328's precedent of building the bundle from the fixture manifest directly and keeping the reducer input a typed bundle that an intake adapter can populate later.

**2. No new fixtures.** The merged corpus already supplies nine scenarios, eighteen artifacts, and fourteen physical files, and the fixture-contract test pins that matrix, so adding fixtures would break `site_core_uses_canonical_rotation_and_coverage_contracts`. The corpus has no retry scenario distinct from backlog; both reduce to `blockedOrDeferred`.

**3. `LikelyContributor` is never produced.** No corpus scenario defines it, and the issue's own vocabulary maps onto symptom, confirmedFailure, blockedOrDeferred, and insufficientEvidence.

**4. Correlation keys are the typed `SccmSiteCoreTransactionKey`, not shared `SccmCorrelationKey` entries.** The spine caps key confidence at `Low` unless the key names a registered stable profile, and that registry is deliberately empty. Lane #328 made the same choice.

**5. `hman.log` and `statesys.log` are coverage-visible but not fact-bearing.** They are declared members of the site-core groups and their capture states survive, which the `incomplete-statesys-absent` gap depends on, but no fixture validates the component name their records carry. Guessing `SMS_HIERARCHY_MANAGER` or `SMS_STATE_SYSTEM` would be unverified, so they report `unsupported` coverage instead. Commit `05f87450` exists specifically to close the alternative, which was silently contributing nothing.

**6. Exported status-system messages and site configuration facts** (the optional half of the issue's input contract) are not consumed; no corpus fixture declares them.

**7. Two implemented branches are unexercised by the corpus** and are covered by focused tests instead: the same-instant same-phase conflicting-outcome guard (which falls to insufficient evidence rather than letting vector order pick a winner) and the moderate confidence ceiling for a transaction with no ComponentStart entry evidence.

## Corpus inconsistency worth a decision

`rotation-boundary/expected.json` omits `completeLogicalRecord` on its coverage-gap evidence while `malformed` and `incomplete` declare `false` on theirs, and no rule distinguishes the three (all cite a synthesized physical tail reference). The reducer emits the flag uniformly; the contract test compares gap evidence without it and separately asserts the stricter invariant that every coverage-gap evidence carries `completeLogicalRecord == false`. Aligning the corpus would let that carve-out be deleted.

Refs #327. No native Windows acceptance is claimed; this is pure parser-layer work.

## Summary by CodeRabbit

* **New Features**
  * Added analysis support for SCCM site-core server evidence.
  * Reports validated component, queue, status, and terminal-state results.
  * Provides confidence levels, observations, findings, coverage gaps, and targeted artifact recapture requests.
  * Handles incomplete, conflicting, malformed, capped, rotated, and recovered evidence while limiting unsupported conclusions.
  * Produces consistent, deterministic results across supported evidence bundles.

* **Tests**
  * Added comprehensive validation for evidence integrity, source compatibility, chronology, confidence handling, request limits, and result consistency.
  * Added performance coverage for large evidence sets.

### DIFF
--- a/crates/cmtraceopen-parser/src/sccm/server/windows/mod.rs
+++ b/crates/cmtraceopen-parser/src/sccm/server/windows/mod.rs
@@ -1,7 +1,9 @@
 mod catalog;
 mod intake;
 mod management_point;
+mod site_core;
 
 pub use catalog::*;
 pub use intake::*;
 pub use management_point::*;
+pub use site_core::*;
--- a/crates/cmtraceopen-parser/src/sccm/server/windows/site_core.rs
+++ b/crates/cmtraceopen-parser/src/sccm/server/windows/site_core.rs
@@ -1,0 +1,1614 @@
+//! Server-local site-core and status-system analysis for issue #327.
+//!
+//! The reducer owns evidence produced by the site server itself: the site
+//! component manager family (`server-sitecomp`) and the status/state system
+//! family (`server-status`). It reduces complete logical records into
+//! component-keyed transactions along
+//! `ComponentStart -> ComponentWork -> InboxOrQueue -> StatusOrStateProcessing
+//! -> HealthyOrTerminal`.
+//!
+//! Scope boundaries this module never crosses:
+//!
+//! * Management Point, Distribution Point, Software Update Point, hierarchy
+//!   transport and Provider/Admin Service evidence belongs to other lanes.
+//!   `mpcontrol.log` is site-server produced but carries a Management Point
+//!   workflow subject, so it stays with the `server-mp-policy` source and is
+//!   never admitted here.
+//! * No client impact, downstream-role availability or cross-side causal
+//!   statement is derivable from this analyzer; those claims are declared
+//!   prohibited on every result.
+//!
+//! Only the `sccm-site-core` v1 extraction profile is selected, and only for
+//! the synthetic `5.00.TEST` corpus version. A fact requires a complete logical
+//! record, an exact profile-validated component and work-item key, the
+//! declared producer component for its source family, and a source whose
+//! catalogued basename, family and rotation all agree with its manifest
+//! declaration.
+
+use std::cmp::Ordering;
+use std::collections::{BTreeMap, BTreeSet};
+
+use serde::Serialize;
+
+use crate::models::log_entry::Severity;
+use crate::sccm::{
+    classify_artifact_name, SccmArtifact, SccmArtifactFamily, SccmArtifactRequest, SccmConfidence,
+    SccmCoverageState, SccmEvidence, SccmEvidenceRef, SccmFinding, SccmFindingBuilder,
+    SccmFindingClass, SccmFindingCoverageGap, SccmPhase, SccmRole, SccmRotation,
+    SccmTerminalEvidence, SccmTimeOrderingState, SccmTimestamp,
+};
+
+pub const SCCM_SITE_CORE_ANALYSIS_SCHEMA_VERSION: u32 = 1;
+pub const SCCM_SITE_CORE_PROFILE_ID: &str = "sccm-site-core";
+pub const SCCM_SITE_CORE_PROFILE_VERSION: u32 = 1;
+pub const SCCM_SITE_CORE_PROFILE_STABILITY: &str = "experimental";
+pub const SCCM_SITE_CORE_COMPONENT_GROUP: &str = "server-sitecomp";
+pub const SCCM_SITE_CORE_STATUS_GROUP: &str = "server-status";
+
+/// The only ConfigMgr build this profile is validated against.
+const SITE_CORE_PROFILE_VERSION_TOKEN: &str = "5.00.TEST";
+/// Producer component recorded by each admitted source family.
+const COMPONENT_PRODUCER: &str = "SMS_SITE_COMPONENT_MANAGER";
+const STATUS_PRODUCER: &str = "SMS_STATUS_MANAGER";
+/// Floor for a bounded recapture request after a capture limit truncated a
+/// source. Requests never ask for an unbounded file.
+const RECAPTURE_FLOOR_BYTES: u64 = 4096;
+
+const STATE_CHAIN: [SccmSiteCorePhase; 5] = [
+    SccmSiteCorePhase::ComponentStart,
+    SccmSiteCorePhase::ComponentWork,
+    SccmSiteCorePhase::InboxOrQueue,
+    SccmSiteCorePhase::StatusOrStateProcessing,
+    SccmSiteCorePhase::HealthyOrTerminal,
+];
+
+const PROHIBITED_CLAIMS: [SccmSiteCoreProhibitedClaim; 3] = [
+    SccmSiteCoreProhibitedClaim::AbsentDownstreamRole,
+    SccmSiteCoreProhibitedClaim::ClientImpact,
+    SccmSiteCoreProhibitedClaim::CrossSideCausality,
+];
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SccmSiteCoreTopology {
+    pub site_code: String,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct SccmSiteCoreSource {
+    pub artifact: SccmArtifact,
+    pub source_group: String,
+    pub rotation_lineage: String,
+    pub fragment_complete: Option<bool>,
+    pub physical_line_end: Option<u32>,
+    pub capture_limit_bytes: Option<u64>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct SccmSiteCoreBundle {
+    pub topology: SccmSiteCoreTopology,
+    pub sources: Vec<SccmSiteCoreSource>,
+    pub evidence: Vec<SccmEvidence>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum SccmSiteCoreWorkflow {
+    SiteCore,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum SccmSiteCorePhase {
+    ComponentStart,
+    ComponentWork,
+    InboxOrQueue,
+    StatusOrStateProcessing,
+    HealthyOrTerminal,
+}
+
+impl SccmSiteCorePhase {
+    fn serialized_name(self) -> &'static str {
+        match self {
+            Self::ComponentStart => "componentStart",
+            Self::ComponentWork => "componentWork",
+            Self::InboxOrQueue => "inboxOrQueue",
+            Self::StatusOrStateProcessing => "statusOrStateProcessing",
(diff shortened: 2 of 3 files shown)


adamgell:
@coderabbitai full review

adamgell:
@coderabbitai full review

adamgell:
### Code review
Five independent reviewers examined this change from different angles (CLAUDE.md compliance, focused bug scan, git-blame history, prior-PR feedback, and code-comment guidance). Each candidate finding was then independently verified and confidence-scored; anything below 80 is reported separately as unconfirmed.
**5 confirmed finding(s)** (score >= 80), 21 unconfirmed, 0 rejected as false positives.

#### 1. Result advertises a findingClass whose finding is silently discarded by `.ok()?`
`cmtraceopen-parser/src/sccm/server/windows/site_core.rs` lines 845-864, 1328-1383 (drop at 1376) (confidence 85)
Reproduced exactly. Built a bundle containing only the healthy fixture's sitecomp.log (statmgr.log simply not collected, so no source row and no gap). Output: results[0].findingClass = "insufficientEvidence", results[0].state = "incomplete", coverageGaps = [], findings = []. The chain is exactly as described: the else branch at 824-831 sets Some(InsufficientEvidence); transaction_gap_ids is empty; SccmFinding::validate hits findings.rs:412 MissingCoverageGap; .ok()? at 1376 swallows it. The analysis self-contradicts and the PR's own invariant test every_classified_outcome_carries_a_validated_shared_finding would fail on this bundle. Reachability is not exotic: a transaction with sitecomp records and no status terminal record and fully-captured sources is the ordinary mid-work case. Verified error swallowing also exists at 1411.

Repro output: results[0] = {state: "incomplete", findingClass: "insufficientEvidence", confidence: "none", coverageGapArtifactIds: [], nextArtifacts:[server-status request]}; findings = []; coverageGaps = [].

#### 2. Quadratic evidence-uniqueness scan makes the reducer unusable on real log volumes
`cmtraceopen-parser/src/sccm/server/windows/site_core.rs` lines 1570-1582, called at 649 (confidence 80)
Independently measured, and my numbers match the candidate's almost exactly. Release build, analyze_site_core only: 500=7.7ms, 1k=15.6ms, 2k=30.9ms, 4k=88.6ms, 8k=321.8ms, 16k=1.03s, 32k=4.33s. I then isolated it with a single-workItem bundle (one transaction, so per-transaction work is O(1) groups) and the curve is unchanged: 32k=4.72s. That pins the growth on the per-evidence admission scan, exactly as claimed: .take(2) only short-circuits on the rare duplicate, so the unique path walks all of bundle.evidence per record. Default 2.5 MB ConfigMgr server logs are roughly 10k records each, and a sitecomp+statmgr+rotation bundle is comfortably 30-60k, i.e. seconds to tens of seconds. The proposed O(E log E) pre-pass preserves the semantics. Real, measured, and structural.

Measured twice; second run isolates the quadratic to evidence_identity_is_unique rather than per-transaction reduction.

#### 3. Fragment observations bypass the validated-producer gate, so an unsupported source family gets a Symptom diagnosis
`cmtraceopen-parser/src/sccm/server/windows/site_core.rs` lines 1026-1045 (with 1098-1105 and 325-333) (confidence 82)
Reproduced end to end. A statesys.log source (admitted by shape, producer: None) with an unparsed tail yields all three predicted outputs simultaneously: unlinkedObservations = [{observationId: "malformed-status-record", findingClass: "symptom", confidence: "low"}], findings[].coverageGaps = [{coverage: "parseFailed"}], and coverageGaps = [{state: "unsupported", no evidence}] for the same artifact id. Critically, results[0].coverageGapArtifactIds = [] because 329-333 removed the fragment artifact, which breaks the invariant the PR's own test asserts at tests/sccm_server_site_core.rs:509-513. collect_fragments genuinely has no producer check while source_carries_facts (662) and collect_coverage_gaps (1102) both do. This directly undercuts the head commit's stated purpose, and the one-line filter fix is correct.

Repro Case E. Also confirmed the observation id is keyed on group (format!("malformed-{}-record", group.slug())), so an unvalidated statesys fragment would merge into the same Symptom finding as a validated statmgr fragment.

#### 4. Producer gate added in 05f87450 skips collect_fragments, so an unvalidated family still emits a Symptom finding and contradicts its own coverage gap
`cmtraceopen-parser/src/sccm/server/windows/site_core.rs` lines 1026-1045 (collect_fragments), 661-670 (source_carries_facts), 1098-1106 (collect_coverage_gaps) (confidence 80)
Duplicate of the collect_fragments producer-gate finding, framed against the head commit, and fully reproduced. All three predicted outputs appear together for one statesys.log artifact: a Symptom/Low malformed-status-record observation and finding, a hardcoded ParseFailed coverage gap inside that finding, and a top-level Unsupported gap with no evidence. The head commit's message does state its purpose as reporting an unvalidated family as unsupported coverage, and the gate landed at two of the three consumers. The group-keyed observation id merging an unvalidated source into the same Symptom finding as a validated one is also correct. The proposed repro path via the malformed fixture is sound; I confirmed the equivalent by constructing the tail directly.

Repro Case E. Marginally below the sibling report only because I did not run the exact malformed-fixture relabel it proposes; the equivalent construction reproduces every claimed output.

#### 5. A classified transaction can silently lose its shared finding (`.ok()?` swallows a contract violation)
`cmtraceopen-parser/src/sccm/server/windows/site_core.rs` lines 845-864, 1328-1383 (esp. 1376) (confidence 85)
Duplicate of the first finding with a more precise reachability trace, and it is exactly right. I confirmed the trace line by line: the else branch at 824-831 sets Incomplete + Some(InsufficientEvidence); transaction_gap_ids at 329-333 is empty because collect_coverage_gaps produced nothing; build_finding constructs gaps = []; findings.rs:412-414 returns MissingCoverageGap; .ok()? at 1376 discards it. Reproduced with the equivalent bundle: results[0].findingClass = "insufficientEvidence" with findings = []. The observation that the same swallow exists at 1411 for observations is also correct. Both remedies proposed (propagate the error, or guarantee an Incomplete transaction always carries a gap) are sound.

Repro Case A confirms the emitted analysis declares findingClass "insufficientEvidence" with no matching entry in findings.

---

<details><summary>Unconfirmed findings (scored 50-79, verify before acting)</summary>

- **[60]** `healthy` at `high` confidence over an unparsed tail in the transaction's own source
- **[68]** Unorderable failure-vs-recovery contradiction resolves to confirmedFailure instead of insufficient evidence
- **[70]** `unsupported` coverage gap is relabelled `parseFailed` inside the finding that cites it
- **[50]** `next_power_of_two` on a caller-supplied u64 panics in debug and silently degrades in release
- **[65]** rotation_rank collapses distinct Numbered/Timestamped rotations, so a real rotation boundary is misclassified as a malformed record
- **[62]** truncated_tail measures parse coverage from unadmitted evidence, so rejected records vanish without a coverage gap
- **[65]** BlockedOrDeferred never clears: a later success in the same phase does not supersede an earlier deferral
- **[70]** finding_gap relabels unsupported coverage as parseFailed inside the emitted findings
- **[66]** Rotation-boundary observation hardcodes the component entry phase for status-family fragments
- **[75]** Fact collection is O(evidence^2) over the whole bundle
- **[70]** rotation_rank repurposed as rotation identity collapses distinct numbered/timestamped rotations, misclassifying a rollover as a source anomaly
- **[75]** Fact admission is O(N^2) over the whole evidence vector with no early exit on the common path
- **[70]** Next-artifact request names the status source that provably cannot yield a fact, and never names statmgr.log
- **[68]** Interior unparsed regions are invisible to coverage, so a captured source with a hole still reports healthy at high confidence
- **[55]** Evidence quarantined for identity collision is dropped with no coverage gap, finding, or request
- **[50]** Caller-supplied capture_limit_bytes can panic the reducer via next_power_of_two overflow
- **[72]** A captured source flagged `fragmentComplete: false` with no unparsed tail vanishes from the analysis entirely
- **[72]** The status re-collection request asks for `statesys.log`, a source the profile has just declared unreadable, and cross-products rotated filenames
- **[50]** `next_power_of_two` on an unchecked capture limit panics in debug / silently shrinks the request in release
- **[75]** Evidence admission is quadratic in the number of parsed records

</details>

adamgell:
@coderabbitai full review

adamgell:
Superseded by corrected draft #439: https://github.com/adamgell/cmtraceopen/pull/439

Do not merge this draft. Closing #405 without merge in favor of the independently reviewed symmetric fix lineage. Issue #327 intentionally remains open.

PR #408, opened 2026-08-01
fix(ccm): validate signless time-tail offsets

### LINKED ISSUE #406: CCM parser fabricates a UTC offset from microsecond timestamps
Found by manual code review of the merged parser.

`split_ccm_time_tail` in `crates/cmtraceopen-parser/src/parser/ccm.rs` (around line 188) contains:

```rust
if value.len() == 6 {
    return (&value[..3], Some(&value[3..]));
}
```

Any signless six-digit fractional tail is therefore read as three millisecond digits plus three offset digits. A microsecond-precision timestamp `<time="10:00:00.123456">` parses as 123 ms with a **+456 minute offset**, silently shifting the resulting UTC by 7 hours 36 minutes, and the result is marked `NormalizedUtc`, the highest-confidence ordering state. The function's own doc comment explicitly exempts the seven-digit precision emitted by IME logs but says nothing about six.

This is a wrong-answer path rather than a display issue, and it is in shared parser code reached by the SCCM lanes, the ESP/Intune pipelines, and the general log viewer.

**Why the naive fix is wrong.** The signless six-digit form is genuinely ambiguous. The documented legacy CCM grammar is `%03u%d`, and `%d` prints a positive offset without a sign, so `.123480` really does mean 123 ms with a +480 minute (+8h) offset in real CCM output. Deleting the branch would break legitimate positive-offset timestamps; keeping it fabricates offsets from microseconds.

**Direction.** The program's own constitution already answers this class: use UTC-normalized ordering only when the record has a *valid* offset, and downgrade unvalidated patterns rather than guessing. Validating the candidate against what a real offset can be (|minutes| <= 840, in practice a multiple of 15) separates the two readings: 480 is a real offset and keeps the legacy interpretation, 456 is not and should be read as fractional seconds with no offset. Where a value is plausible under both readings, the choice needs to be stated explicitly and the ordering confidence considered, not silently resolved.

A test-first fix is in progress on `codex/ccm-ambiguous-time-tail`, including a blast-radius assessment across the ESP/Intune surfaces and a judgment on whether the parallel public projection `split_legacy_public_time_tail` carries a related defect.

Refs #317.

adamgell:
Fixes #406. A signless six-digit fractional tail was unconditionally read as three millisecond digits plus three offset digits, so `<time="10:00:00.123456">` parsed as 123 ms with a +456 minute offset, shifting UTC by 7h36m and stamping the record `NormalizedUtc`, the highest-confidence ordering state.

## Why the obvious fix is wrong

The form is genuinely ambiguous. The documented legacy grammar is `%03u%d`, and `%d` prints a positive offset unsigned, so `.123480` really does mean 123 ms plus a +480 minute offset in real CCM output. Deleting the branch would break legitimate positive-offset timestamps.

This is also the second time this split has been implemented wrongly. The original greedy regex `(?P<ms>\d+)(?P<tz>[+-]*\d+)` assigned the final digit to the offset, so `.123456` gave ms=12345, tz=6. Commit 3e53bd30 replaced it with the six-digit rule, which gives ms=123, tz=456. Both attempts resolved an ambiguous grammar with a positional heuristic. This change stops doing that.

## The rule

Digit width only nominates candidates; validity decides. A signless tail splits into `%03u%d` only when the trailing run is a real timezone offset:

1. No leading zero, since `%d` never zero-pads, so `123045` and `123000` are fractional text.
2. Magnitude at most 840 minutes (UTC-14 to UTC+14), so `123900` is fractional.
3. A whole quarter hour, since no zone in current use has a non-quarter-hour offset, so `123456` and `123481` are fractional.

`.123480` still yields +480. `.000240` still yields +240. Seven-digit IME precision and every signed tail are untouched, including out-of-range signed values, which stay `Some(99999)` with `OffsetInvalid`: an explicit sign is the source stating its own provenance and is never screened.

## Blast radius

`CcmTimestampParse` reaches exactly one surface, the SCCM evidence spine, via `scan_logical_records` (crate-private, consumed only by `sccm/ingest.rs`). So the fabricated offset corrupted SCCM evidence `offsetMinutes`, `utcMillis`, and `orderingState`, and therefore cross-artifact ordering and every finding built on it. The general viewer, timeline, ESP/Intune ingestion, and tailing all build `LogEntry` from `split_legacy_public_time_tail` instead and are unaffected by this change. IME-specialized Intune logs bypass CCM entirely.

One existing fixture uses a signless six-digit tail: `sccm_spine_contract.rs:4462` with `time="10:00:00.000240"`, asserting `offset_minutes == Some(240)` and `NormalizedUtc`. 240 is a real offset, so it keeps the legacy reading and passes unmodified. No fixture or expectation was edited to make anything pass.

## The frozen table

`ccm_time_tail_ambiguity_table_is_frozen` pins 26 rows: unsigned tails of one to eight digits covering valid and invalid offset candidates, quarter-hour and boundary cases at 105 and 840, zero-padded and over-range values, plus signed counterparts. Each row pins fractional text, milliseconds, offset minutes, ordering state, and `utcMillis`, and asserts `utcMillis.is_some() == (state == NormalizedUtc)`.

`ccm_time_tail_rejects_both_historical_offset_bugs` pins the two historical wrong answers explicitly, so a third attempt cannot silently pick a new split. The doc comment at the function names both prior implementations and the exact wrong answers they produced.

## Honest residual

Of the 1000 possible three-digit trailing runs, 50 pass all three rules, so a microsecond timestamp still has roughly a 5 percent chance of being read as an offset. I kept `NormalizedUtc` for a validated split deliberately: the current enum has no free downgrade (`OffsetMissing` would contradict a populated `offsetMinutes`), so downgrading means a new serialized variant plus frontend churn, and it would strip ordering from every genuine positive-offset CCM record, which is the common case for UTC+N tenants. If that residual is unacceptable, the clean follow-up is an explicit `OffsetAmbiguous` state rather than another split rule, and the frozen table makes that a one-column change.

## Verification

- `cargo test --locked -p cmtraceopen-parser`: 963 passed, 0 failed
- `cargo test --locked --workspace`: 1717 passed, 0 failed
- `cargo clippy --locked -p cmtraceopen-parser --all-targets -- -D warnings`: clean
- `cargo check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown`: clean
- `git diff --check e56ad387 HEAD` clean; `rustfmt --check` clean on both changed files; `npx tsc --noEmit` clean

Refs #317.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CCM timestamp parsing to correctly distinguish timezone offsets from fractional seconds.
  - Validated unsigned timezone offsets against supported bounds and quarter-hour increments.
  - Preserved signed offsets while reporting invalid values appropriately.
  - Fixed handling of six-digit fractional seconds, retaining full microsecond precision without incorrect timezone normalization.
- **Tests**
  - Added comprehensive regression coverage for valid, invalid, signed, and ambiguous timestamp formats.

### DIFF
--- a/crates/cmtraceopen-parser/src/parser/ccm.rs
+++ b/crates/cmtraceopen-parser/src/parser/ccm.rs
@@ -970,4 +970,150 @@ mod tests {
         );
         assert!(records[0].timestamp.utc_millis.is_some());
     }
+
+    fn logical_record_for_time_tail(tail: &str) -> CcmLogicalRecord {
+        let text = format!(
+            concat!(
+                r#"<![LOG[Tail {tail}]LOG]!><time="10:00:00.{tail}" date="07-30-2026" "#,
+                r#"component="PolicyAgent" context="" type="1" thread="42" file="policyagent.cpp">"#
+            ),
+            tail = tail
+        );
+        let mut records = scan_logical_records(&text, "PolicyAgent.log");
+        assert_eq!(records.len(), 1, "tail {tail}: expected one logical record");
+        records.remove(0)
+    }
+
+    /// Frozen answers for every shape of CCM fractional-second tail.
+    ///
+    /// Two shipped implementations resolved this ambiguity by counting digits
+    /// and both were wrong (see `split_ccm_time_tail`). This table is the
+    /// regression barrier: it pins the fractional text, the millisecond value,
+    /// the source offset, and the ordering confidence for unsigned tails of
+    /// one to eight digits and for their signed counterparts. Change a row
+    /// only with a documented grammar reason, never to let a new heuristic
+    /// pass.
+    #[test]
+    fn ccm_time_tail_ambiguity_table_is_frozen() {
+        use CcmTimestampParseState::{NormalizedUtc, OffsetInvalid, OffsetMissing};
+
+        // (time tail, fractional text, milliseconds, source offset, ordering state)
+        let cases: [(&str, &str, u32, Option<i32>, CcmTimestampParseState); 26] = [
+            // Unsigned tails that cannot be `%03u%d`: no trailing run of
+            // digits reads as a real UTC offset, so they stay fractional and
+            // the record is never promoted to UTC-normalized ordering.
+            ("1", "1", 1, None, OffsetMissing),
+            ("12", "12", 12, None, OffsetMissing),
+            ("123", "123", 123, None, OffsetMissing),
+            // `%d` prints a zero offset as "0", so a four-digit tail is not
+            // evidence of one; downgrade instead of guessing.
+            ("1234", "1234", 123, None, OffsetMissing),
+            ("12345", "12345", 123, None, OffsetMissing),
+            // Microsecond precision. 456 is not a UTC offset.
+            ("123456", "123456", 123, None, OffsetMissing),
+            // `%d` never zero-pads, so "045" is fractional text and not +45.
+            ("123045", "123045", 123, None, OffsetMissing),
+            // Same rule: a padded zero offset always arrives signed ("+000").
+            ("123000", "123000", 123, None, OffsetMissing),
+            // 481 minutes is not a whole quarter hour.
+            ("123481", "123481", 123, None, OffsetMissing),
+            // 900 minutes exceeds UTC+14:00.
+            ("123900", "123900", 123, None, OffsetMissing),
+            // Unsigned tails that are genuine `%03u%d` records: three
+            // millisecond digits followed by a positive offset.
+            ("000240", "000", 0, Some(240), NormalizedUtc),
+            ("123480", "123", 123, Some(480), NormalizedUtc),
+            ("123840", "123", 123, Some(840), NormalizedUtc),
+            ("123105", "123", 123, Some(105), NormalizedUtc),
+            // IME writes seven fractional digits, never an offset.
+            ("1234567", "1234567", 123, None, OffsetMissing),
+            ("12345678", "12345678", 123, None, OffsetMissing),
+            // Signed tails are self-delimiting: the sign is the source's own
+            // statement of provenance and is never screened for plausibility.
+            ("123+480", "123", 123, Some(480), NormalizedUtc),
+            ("123-240", "123", 123, Some(-240), NormalizedUtc),
+            ("000+000", "000", 0, Some(0), NormalizedUtc),
+            ("123+481", "123", 123, Some(481), NormalizedUtc),
+            ("123+0", "123", 123, Some(0), NormalizedUtc),
+            ("1+2", "1", 1, Some(2), NormalizedUtc),
+            ("123456+480", "123456", 123, Some(480), NormalizedUtc),
+            ("1234567-060", "1234567", 123, Some(-60), NormalizedUtc),
+            // Out-of-range signed offsets stay reported and stay uncomparable.
+            ("123+99999", "123", 123, Some(99999), OffsetInvalid),
+            ("123-99999", "123", 123, Some(-99999), OffsetInvalid),
+        ];
+
+        for (tail, fraction_text, millis, offset_minutes, ordering_state) in cases {
+            let (split_fraction, split_offset) = split_ccm_time_tail(tail);
+            assert_eq!(
+                split_fraction, fraction_text,
+                "tail {tail}: fractional text"
+            );
+            assert_eq!(
+                split_offset.is_some(),
+                offset_minutes.is_some(),
+                "tail {tail}: offset presence"
+            );
+            assert_eq!(
+                truncate_subsecond_to_millis(split_fraction),
+                Some(millis),
+                "tail {tail}: milliseconds"
+            );
+
+            let record = logical_record_for_time_tail(tail);
+            let timestamp = &record.timestamp;
+            assert_eq!(
+                timestamp.original_display,
+                Some(format!("07-30-2026 10:00:00.{fraction_text}")),
+                "tail {tail}: original display"
+            );
+            assert_eq!(
+                timestamp.offset_minutes, offset_minutes,
+                "tail {tail}: source offset"
+            );
+            assert_eq!(
+                timestamp.ordering_state, ordering_state,
+                "tail {tail}: ordering state"
+            );
+
+            let naive = chrono::NaiveDate::from_ymd_opt(2026, 7, 30)
+                .unwrap()
+                .and_hms_milli_opt(10, 0, 0, millis)
+                .unwrap();
+            let expected_utc =
+                offset_minutes.and_then(|minutes| normalized_utc_millis(naive, minutes));
+            assert_eq!(
+                timestamp.utc_millis, expected_utc,
+                "tail {tail}: normalized utc"
+            );
+            assert_eq!(
+                timestamp.utc_millis.is_some(),
+                ordering_state == NormalizedUtc,
+                "tail {tail}: utc millis must exist exactly when ordering is normalized"
+            );
+        }
+    }
+
+    /// Both shipped attempts fabricated an offset from a microsecond tail.
+    #[test]
+    fn ccm_time_tail_rejects_both_historical_offset_bugs() {
+        let timestamp = logical_record_for_time_tail("123456").timestamp;
+
+        assert_ne!(
+            timestamp.offset_minutes,
+            Some(6),
+            "greedy-regex attempt read the final digit as the offset"
+        );
+        assert_ne!(
+            timestamp.offset_minutes,
+            Some(456),
+            "digit-count attempt read the final three digits as the offset"
+        );
+        assert_eq!(timestamp.offset_minutes, None);
+        assert_eq!(
+            timestamp.ordering_state,
+            CcmTimestampParseState::OffsetMissing
+        );
+        assert_eq!(timestamp.utc_millis, None);
+    }
 }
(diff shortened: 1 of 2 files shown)


adamgell:
@coderabbitai full review

adamgell (commented):
Review of PR #408 (fix(ccm): validate signless time-tail offsets) at exact head eab83fac.

Independent reviewer, fresh detached worktree at eab83fac7ad8af7d3c0b47a7bcdbd8152d5a138f.
Worktree verified clean (git status --porcelain empty) at start and at end. No commits,
pushes, or modifications made. All probe code was built outside the repository.

VERDICT: PASS

BATTERY
  cargo test --locked -p cmtraceopen-parser      963 passed, 0 failed (354+222+8+3+3+47+29
                                                 +12+6+48+28+11+1+30+18+5+138)
  cargo test --locked --workspace               1717 passed, 0 failed
  cargo clippy --locked -p cmtraceopen-parser --all-targets -- -D warnings   clean, exit 0
  cargo check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown clean, exit 0
  git diff --check e56ad387 HEAD                 exit 0, no output
  rustfmt --edition 2021 --check on both changed files                       exit 0

VALIDITY RULE: correct and complete for the shape it governs

I attempted to find a real timezone offset the three-part test wrongly rejects, and could
not. Because the split is gated on a total tail width of 6, the candidate offset is always
exactly three digits, so the reachable candidate set is 100..999.

Every distinct UTC offset in use since 1990 was driven end to end through the public API.
All that the shape can express are accepted with correct utcMillis, including every
irregular one: +330 (India), +345 (Nepal), +765 (Chatham), +210 (Iran), +570 (Marquesas
and ACST), +630 (Lord Howe), +525 (Eucla), +390 (Myanmar), plus +120/+240/+480/+840.
Offsets rejected by the validity test at width 6: zero.

Both bounds hold. No offset in use since 1990 is a non-quarter-hour value, and none
exceeds 840 minutes. The last non-quarter-hour offset predates the SCCM era by decades.

The no-leading-zero rule is correct for every candidate it can actually see. A three-digit
run beginning with 0 cannot be %d output, since %d never pads. Verified: .123045 and
.123090 and .123015 all stay fractional.

The change is a strict narrowing. Across all 1,111,110 signless tails of length 1 to 6,
the number that gain an offset relative to e56ad387 is 0. 950,000 six-digit tails lose a
fabricated offset; 50,000 keep one. Lengths 1 through 5 are 100 percent identical to the
parent. The rule is therefore structurally incapable of producing a new wrong answer of
the family that sank attempts one and two.

RESIDUAL: number confirmed, disposition defensible but under-argued

Independently computed: of the 1000 three-digit runs, exactly 50 pass all three rules,
spanning 105 to 840 in steps of 15. That is 5.000 percent, matching the disclosure.

On keeping NormalizedUtc: the outcome is defensible, because unlike attempts one and two
the residual fires only on input that is genuinely indistinguishable from a real legacy
record. Attempts one and two fired on input that was distinguishable. That is a real
categorical improvement.

However, the stated counter-argument does not survive contact with the corpus. Across
every file in the repository the CCM tail shapes are: 595 DDD+DDD, 21 DDD-DDD, 19
DDDDDDD, 5 DDD+DDDDD, 5 DDD+DDDD, 2 DDDDDD. Every genuine positive offset attested is
signed, so a downgrade confined to validated-signless splits would strip ordering from
none of them. The two six-digit signless occurrences are both in the new test file; the
shape is not attested in any real fixture.

Separately, the cost argument is at its weakest today. There are zero frontend or
TypeScript references to orderingState, no persisted artifacts, and
SCCM_DIAGNOSTICS_SCHEMA_VERSION is still 1. Adding a variant is cheap now and grows more
expensive with every consumer. Recommend either adding the variant or recording the
decision explicitly so the third reader does not relitigate it.

FROZEN TABLE: claims verified

The 26 rows do cover unsigned tails of every length from 1 to 8, valid and invalid
candidates, the boundaries at 105 and 840 (independently confirmed as the true minimum and
maximum accepted values), zero-padded candidates (045, 000), an over-range candidate (900),
a non-quarter-hour candidate (481), and signed counterparts. Each row pins fractional text,
milliseconds, offset minutes, ordering state, and utcMillis, with utcMillis recomputed
independently rather than restated, plus the invariant that utcMillis exists exactly when
ordering is NormalizedUtc. This is a strong barrier.

HISTORICAL BUGS: both pinned

Reimplementing all three rules against .123456 confirms: attempt one yields fraction
\"12345\" offset 6; attempt two yields fraction \"123\" offset 456; HEAD yields fraction
\"123456\" offset None. Both prior answers are now unreachable, and the negative test asserts
that directly rather than only asserting the new answer.

BLAST RADIUS: both claims verified

CcmTimestampParse and CcmTimestampParseState are referenced outside parser/ccm.rs only by
sccm/evidence.rs. scan_logical_records is called only by sccm/ingest.rs (plus tests). Both
are crate-private.

SccmRawEvidenceSnapshot::from_record destructures CcmLogicalRecord and reads only
entry.component, entry.source_file, and entry.message. It never reads entry.timestamp,
entry.timestamp_display, or entry.timezone_offset, so the fallback branch in parse_captures
that consumes split_ccm_time_tail output is unreachable on both paths.

Both public entry points (parse_content and parse_lines) gate CCM emission on
public_compatible, which is determined solely by split_legacy_public_time_tail. That
function is unchanged in this diff.

Confirmed empirically rather than only by reading: the public LogEntry surface was dumped
against both e56ad387 and eab83fac across 1,111,369 tail shapes, covering every signless
tail of length 1 to 6, selected 7 to 10 digit tails, and 252 signed shapes, recording
format, timezone_offset, timestamp, timestamp_display, message, severity, line number,
thread, and the error count. The two dumps are byte identical, md5
24eb36b9087b6ba39f46981a3280e78b. There are no public LogEntry byte changes.

FIXTURE REGRESSION: passes for the right reason

signless_ccm_offset_is_enriched_only_in_sccm_provenance still yields offset 240 at
NormalizedUtc. Verified this is because 240 satisfies all three rules independently (no
leading zero, at or under 840, divisible by 15), not because of any width coincidence. The
neighbouring pre-spine baseline test also still passes on 000, 123, 000240, and 1234567.

GAPS (none blocking; item 1 is the one I would want tracked)

1. Pre-existing panic on non-ASCII decimal digits, on the line this PR wrote. The regex
   crate's \\d is Unicode-aware by default, so time_tail can capture multi-byte digits while
   h, m, and s remain ASCII and parse successfully. The byte-indexed slices then panic on a
   char boundary. Reproduced at HEAD ccm.rs:167 (truncate_subsecond_to_millis), :228
   (split_ccm_time_tail, the changed line) and :251 (split_legacy_public_time_tail) with
   tails such as .١٢٣ (three Arabic-Indic digits, six bytes), .123١, .12١٢, and .1234١.
   Confirmed identical at e56ad387 (:167, :189, :211), so this PR does not introduce it.
   Raised here because the PR rewrites one of the three sites and its test claims to freeze
   \"every shape of CCM fractional-second tail\", which holds only for ASCII. A
   value.is_ascii() guard plus one table row would close it. The crate is published and
   documented as pure with no I/O, so a panic on untrusted input is worth a follow-up
   issue.

2. Doc and implementation disagree about scope. The doc comment says the split happens when
   \"the trailing run\" is a real offset, but the implementation only ever tests a three-digit
   run at width 6. signless_offset_is_real(\"60\") returns true yet is unreachable, and
   signless_offset_is_real(\"0\") returns false even though \"0\" is exactly what %d prints for
   a zero offset. Harmless today because the width gate makes both unreachable, but it is a
   trap for anyone who later widens the gate. Worth narrowing the comment to describe the
   three-digit case, or moving the width condition inside the helper.

3. Consequence of that gate, pre-existing and unchanged. A legacy writer at offset 0 emits
   .1230 and at offset 60 emits .12360. Neither is six wide, so neither is considered, and
   both land at OffsetMissing. Verified byte identical to the parent, so not a regression,
   and the millisecond value stays correct in both cases. Noting it so the limitation is on
   the record rather than rediscovered as a fourth bug.

4. Table gap: signed offsets beyond i32. .123+2147483648 and .123456-99999999999 produce
   orderingState offsetInvalid with offsetMinutes null, and demote the public entry to Plain
   with errors=1. This contradicts the row comment \"Out-of-range signed offsets stay
   reported and stay uncomparable\", which holds only within i32. The behavior is
   fail-closed and acceptable; the claim is what needs adjusting, or a row added.

5. Minor table gaps, all verified correct in behavior: .123100 (in range, not a quarter
   hour, at the low end), .123855 (over range and not a quarter hour simultaneously),
   .000000 (all zeros), and tails of 9 or more digits.

6. Adjacent, out of scope: crates/cmtraceopen-parser/src/parser/cmtlog.rs:40 still carries
   attempt one's exact greedy regex, (?P<ms>\\d+)(?P<tz>[+-]*\\d+). It does not reach the
   SCCM spine, since scan_logical_records is CCM-only, but it is the same defect family in
   a public LogEntry producer. Worth a separate issue.

7. See the residual section above regarding the ordering-confidence argument.

SUMMARY

This is the first of the three attempts that does not resolve the ambiguity positionally,
and it is verifiably safe in the direction that matters: it cannot fabricate an offset that
the previous implementation did not already fabricate, it rejects no real timezone offset
it is capable of seeing, and it leaves the public LogEntry surface byte for byte identical.
The frozen table and the negative test together make a fourth regression of this family
difficult to land accidentally. Recommend merge, with item 1 filed as a follow-up.

PR #392, opened 2026-07-31
feat(sccm): analyze client health and location evidence

adamgell:
## Summary

Production implementation for #320, stacked temporarily on #391 only to reuse the pure normalized SCCM bundle input while keeping this issue diff isolated.

- evidence-first Setup → Service → Identity → SiteAssignment → ManagementPoint → Transport reducer
- existing nine sanitized scenarios plus adversarial fail-closed coverage
- no policy-output dependency, native collection, live SCCM claim, or server/MP causal diagnosis

## Current TDD state

RED commit `397dca6`: `cargo test --locked -p cmtraceopen-parser --test sccm_client_health` fails because `analyze_client_health` does not yet exist. Implementation and full gates follow in later issue-scoped commits.

Closes no issue until production behavior, exact-head review, CodeRabbit, and required validation gates are complete.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive SCCM client-health analysis.
  - Reports health phases, findings, recovery status, coverage gaps, and requested diagnostic artifacts.
  - Correlates setup, service, identity, site-assignment, management-point, and transport evidence.
  - Distinguishes confirmed issues, insufficient evidence, contradictions, and terminal failures.

- **Bug Fixes**
  - Improved handling of malformed, incomplete, unrelated, or conflicting health evidence.
  - Prevents unsupported evidence from producing misleading success or high-confidence failure results.
  - Ensures consistent ordering and duplicate-free analysis output.

PR #420, opened 2026-08-01
fix(sccm): reject overlapping evidence ranges in the spine

### LINKED ISSUE #418: Spine evidence-reference validation misses overlapping ranges under distinct entry ids
Found during the #321 policy correction round, which fixed the same class **inside its own reducer** — this is the shared-spine half that no lane has closed.

`findings.rs::validate_all_evidence_references` keys on `(artifact_id, entry_id)` and rejects only the *same identity carrying a different range*. Two references with **overlapping line ranges under different `entry_id`s** inside one finding pass validation.

That matters because overlapping ranges mean two references claim the same physical records, which is precisely what makes a citation double-count and what makes an ordering comparator pick arbitrarily. Two reducers have now independently had to fix this in their own code:

- **#328 (Management Point)** fixed it in `management_point.rs` (`evidence_references_overlap`, around lines 1130-1144), after an independent review rejected identical-endpoint-only detection as a P1.
- **#321 (client policy)** just fixed it again in `policy.rs` by copying that same helper, because the class reappeared there.

Each lane fixed its own reducer. The spine validator that every lane's findings pass through still accepts the shape, so a third reducer will hit it, and a finding assembled by any means other than those two reducers can still carry overlapping citations.

**Suggested fix:** extend `validate_all_evidence_references` to reject overlapping ranges within an artifact regardless of `entry_id`, reusing the overlap predicate the two reducers already agree on rather than adding a third copy. Note the two existing copies are themselves duplication worth collapsing into the spine.

Worth pairing with the standing lesson this repository keeps relearning: a defect fixed in one lane is not fixed program-wide until the shared surface enforces it.

Refs #321, #328, #317.

adamgell:
Closes #418.

`findings.rs::validate_all_evidence_references` keyed on `(artifact_id, entry_id)` and rejected only the same identity carrying a different range. Two references with overlapping line ranges under different `entry_id`s inside one finding passed validation, so one physical record could be cited twice and `compare_evidence_refs` would rank the two on span width alone.

## Semantics chosen, and why

The predicate now lives in the spine as `evidence_references_overlap`:

- **Scoped to one artifact.** Line numbers only mean something relative to a source.
- **Inclusive bounds.** `line_start`/`line_end` are inclusive line numbers, so `4-6` and `6-9` share line 6 and overlap, while `1-2` and `3-4` abut and do not. Equal spans are the degenerate overlap and are rejected.
- **A reference carrying no bounds asserts no extent** and therefore cannot be shown to claim another reference's lines. `(None, None)` on either side is not an overlap. This is what keeps existing citations that omit line numbers valid.
- **Inverted ranges are handled by ordering, not by the predicate.** An inverted range reads as empty under an inclusive test and would be silently judged disjoint from everything, which is the failure mode this repo has hit before. The spine closes it by running the validity gate first: every reference clears `validate_evidence_reference` (both bounds present or both absent, `start > 0`, `end >= start`) inside the same function before any pair reaches the overlap sweep, so `InvalidEvidenceReference` still wins and an inverted range can never be mistaken for an empty span.

The deliberate choice not to make the predicate itself fail closed on inverted input is load bearing for the duplication collapse. `management_point.rs` calls it from `evidence_identity_is_unique`, which compares a candidate against every bundle entry, including entries that never passed the reducer's own `safe_evidence_reference` gate. A fail-closed predicate would let one malformed bundle entry suppress good evidence there. Validity is a separate question from extent, and each caller already answers it.

Precedence inside `validate_all_evidence_references` is now: `InvalidEvidenceReference` (per reference) -> `ConflictingEvidenceReference` (one identity, two ranges) -> `OverlappingEvidenceReference` (two identities, one physical line). The first two are unchanged.

The sweep is `O(n log n)`, not pairwise: references are keyed by identity so each entry id contributes exactly one span, spans are sorted per artifact by start line, and a span that clears the widest span seen so far clears every earlier one. `SccmFinding` deserialization has no cap on evidence count, so a quadratic validator on the wire path was not acceptable.

## Did the two reducer copies agree?

Yes, byte-identical. Verified by extracting both and diffing:

- `src/sccm/server/windows/management_point.rs` lines 1130-1144 (merged, #328)
- `src/sccm/client/policy.rs` lines 755-769 on `origin/codex/sccm-321-policy-analysis` (in flight, #391)

Both produced the same text for artifact scoping, inclusive-bounds comparison, and missing-bound handling. No edge case disagreed, so there was no semantics dispute to resolve, and the spine predicate is that text with the reasoning written down.

## Duplication collapse

The management-point copy is deleted; the reducer imports the spine predicate. The removed body is byte-identical to the spine one, so admission there is unchanged, and `sccm_server_management_point` passes unmodified (26/26).

`policy.rs` does not exist on `codex/parser-family-skeleton` yet, so its copy cannot be collapsed here. **#391 should drop `evidence_references_overlap` from `policy.rs` when it rebases** and import `crate::sccm::findings::evidence_references_overlap` instead. `quarantine_overlapping_evidence` stays where it is: it is a reducer policy (quarantine every participant rather than reject the bundle), not the predicate.

## Shipped-fixture audit

Every `expected.json` in the tree was scanned for two references with distinct `entryId` and overlapping ranges inside one artifact, across all three citation surfaces (`evidence`, `terminalEvidence[].reference`, `correlationKeys[].evidence`). The same scan was run against every in-flight SCCM lane branch, since this rejection lands under work already in review.

| Branch | findings | references | violations |
| --- | --- | --- | --- |
| `codex/parser-family-skeleton` (065b8cb7) | 57 | 79 | 0 |
| `codex/sccm-321-policy-analysis` (#391) | 24 | 28 | 0 |
| `codex/sccm-320-health-analysis` (#392) | 24 | 28 | 0 |
| `codex/sccm-319-pure-intake` (#394) | 24 | 28 | 0 |
| `codex/sccm-spine-findings-hardening` (#404) | 24 | 28 | 0 |
| `codex/sccm-327-site-core-reducer` (#405) | 24 | 28 | 0 |
| `codex/sccm-322-deployment-reducer` (#407) | 24 | 28 | 0 |

No shipped fixture carries the shape. Reducer output is covered by the fixture contract suites: they run each reducer over its corpus and diff against `expected.json`, and management point in particular does `builder.build().ok()?`, so a newly rejected finding would silently vanish and fail the comparison. All of them stay green, so no reducer emits an overlapping citation today either.

**One test fixture did carry the shape**, and it is reported rather than papered over. `finding_rejects_key_or_terminal_refs_that_are_not_cited` built its uncited reference with `finding_evidence_ref("client-policy-agent", "policy:2-2")`, and that helper pins every span to line 1 regardless of the entry id. So `policy:2-2` claimed line 1 alongside `policy:1-1`, which is exactly the defect: two entry ids over one physical line, with entry ids that advertise ranges contradicting their own bounds. The fix spells the span out as `2-2` to match the entry id it advertises. The test's subject, that an uncited reference is rejected, is unchanged and still asserts `CorrelationKeyEvidenceNotCited` / `TerminalEvidenceNotCited`. This was a latent inconsistency in the fixture, not a semantics disagreement.

## RED / GREEN

- RED `19addcdb` added the contract plus the inert `OverlappingEvidenceReference` variant so it compiles, and failed with the spine accepting the shape: `called Result::unwrap_err() on an Ok value: SccmFinding { ... evidence: [SccmEvidenceRef { artifact_id: "artifact-a", entry_id: "entry-a", line_start: Some(4), line_end: Some(6) }, SccmEvidenceRef { artifact_id: "artifact-a", entry_id: "entry-b", line_start: Some(4), line_end: Some(6) }] }`. 139 passed, 2 failed.
- GREEN `8b487a35` wired the predicate into the validator and collapsed the management-point copy. 141 passed, 0 failed.

New coverage goes through the real `SccmFinding::validate` path, not a unit helper: builder, direct `validate()`, `serde` deserialization, and the validating serializer. Six overlap shapes (identical, shared start line, shared end line, contained, containing, straddling) are each exercised on all three citation surfaces, and the shapes that must keep validating are pinned too: adjacent spans, equal spans in different artifacts, one bounded plus one unbounded, and both unbounded.

## Verification

All from the worktree, all clean.

- `cargo test --locked -p cmtraceopen-parser --test sccm_spine_contract`: **141 passed, 0 failed** (138 before).
- `cargo test --locked -p cmtraceopen-parser`: **1039 passed, 0 failed** across 21 binaries. Baseline on 065b8cb7 was 1036; the delta is exactly the 3 new spine tests, every other binary count is identical.
- `cargo test --locked --workspace`: **1793 passed, 0 failed** across 36 binaries.
- `cargo clippy --locked -p cmtraceopen-parser --all-targets -- -D warnings`: clean.
- `cargo check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown`: clean.
- `npx tsc --noEmit`: clean.
- `git diff --check 065b8cb7 HEAD`: clean.
- `rustfmt --check` on all three changed files: clean.

Every SCCM contract suite in the tree, run individually, all green:

| Suite | Tests |
| --- | --- |
| `sccm_client_deployment_fixture_contract` | 8 |
| `sccm_client_health_fixture_contract` | 3 |
| `sccm_client_intake_fixture_contract` | 3 |
| `sccm_client_inventory_compliance_metering_fixture_contract` | 47 |
| `sccm_client_management_fixture_contract` | 29 |
| `sccm_client_task_sequence_fixture_contract` | 30 |
| `sccm_client_updates_fixture_contract` | 17 |
| `sccm_correlation_contract` | 12 |
| `sccm_server_advanced_roles_catalog` | 6 |
| `sccm_server_distribution_point_fixture_contract` | 48 |
| `sccm_server_hierarchy_and_replication_fixture_contract` | 28 |
| `sccm_server_intake` | 11 |
| `sccm_server_intake_fixture_contract` | 1 |
| `sccm_server_management_point` | 26 |
| `sccm_server_provider_and_admin_service_fixture_contract` | 30 |
| `sccm_server_software_update_point_fixture_contract` | 18 |
| `sccm_site_core_fixture_contract` | 5 |
| `sccm_spine_contract` | 141 |
| **Total** | **463** |

## Reviewer attention

`SccmFindingValidationError` gains a variant. It is `#[derive(Debug, Clone, Copy, PartialEq, Eq)]` with no `#[non_exhaustive]`, and nothing in the tree matches on it exhaustively, but any lane holding a match should be checked on rebase.

## Summary by CodeRabbit

- **Bug Fixes**
  - Evidence validation now detects and rejects overlapping bounded references for the same artifact and entry.
  - References from primary, terminal, and correlation-key evidence are validated consistently.
  - Valid disjoint, adjacent, cross-artifact, and unbounded references continue to be accepted.
  - Clear validation errors are reported when evidence references conflict.

- **Tests**
  - Expanded coverage for validation, serialization, deserialization, citation ranges, and evidence reference handling.

### DIFF
--- a/crates/cmtraceopen-parser/src/sccm/findings.rs
+++ b/crates/cmtraceopen-parser/src/sccm/findings.rs
@@ -457,6 +457,7 @@ pub enum SccmFindingValidationError {
     InvalidRole,
     InvalidEvidenceReference,
     ConflictingEvidenceReference,
+    OverlappingEvidenceReference,
     MissingEvidenceOrCoverageGap,
     MissingTerminalEvidence,
     InvalidTerminalEvidence,
--- a/crates/cmtraceopen-parser/src/sccm/findings.rs
+++ b/crates/cmtraceopen-parser/src/sccm/findings.rs
@@ -645,11 +645,9 @@ fn validate_roles(finding: &SccmFinding) -> Result<(), SccmFindingValidationErro
     Ok(())
 }
 
-fn validate_all_evidence_references(
-    finding: &SccmFinding,
-) -> Result<(), SccmFindingValidationError> {
-    let mut ranges_by_identity = BTreeMap::new();
-    for reference in finding
+/// Every reference a finding cites, across all three citation surfaces.
+fn cited_evidence_references(finding: &SccmFinding) -> impl Iterator<Item = &SccmEvidenceRef> {
+    finding
         .evidence
         .iter()
         .chain(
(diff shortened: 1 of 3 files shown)


adamgell:
@coderabbitai full review

adamgell:
Review follow-up and CI-base sync — exact head 3615a87de51fd1ce1eb8cca10c44aa9bce30647a

I independently checked the CodeRabbit three-span nitpick before changing the test. The proposed 1-9, 2-3, 5-6 case cannot exercise retained-widest behavior: after sorting, 1-9 and 2-3 overlap, so this validator returns immediately and never inspects 5-6. More generally, because the function exits on the first overlap, the widest update occurs only after a disjoint interval; with start-sorted intervals that current interval is then also the adjacent interval with the furthest end. Adding the proposed case would pass but would not prove the stated branch property, so I did not add a misleading test.

I non-force merged the updated integration base solely to activate PR #429 hosted CI. The feature delta remains the same three files. Fresh local gates on the synchronized head: spine 141/141, wasm32 check green, strict parser Clippy green, scoped rustfmt green, and diff check green.

@coderabbitai review

Please review exact head 3615a87de51fd1ce1eb8cca10c44aa9bce30647a and re-evaluate the three-span rationale above. A no-finding approval must be attached to this head before merge.

adamgell:
Reviewed Copilots suppressed semver note at exact 3615a87d. It is not a blocker for this integration-branch repair: origin/main has no sccm/findings.rs and therefore no released SccmFindingValidationError contract; the new variant is being added within the still-unintegrated SCCM API review sequence, not to an already-published mainline enum. Public LogEntry is unchanged. If the SCCM error type later needs extension after mainline release, non-exhaustive handling can be evaluated as an intentional public-API change rather than silently widening this overlap-fix slice.

adamgell:
Exact integration resync checkpoint — `c7ed2d04d4e5bffa2d942fa18b2c0625c42193c7`

Merged current `origin/codex/parser-family-skeleton` non-force after PR #431 landed. The net feature delta remains exactly the intended three files.

Fresh local verification on this exact head:
- SCCM spine contract: 141/141
- Management Point contract: 26/26
- full parser tests: green
- `wasm32-unknown-unknown` check: green
- strict parser Clippy: green
- scoped rustfmt and `git diff --check`: green
- local CodeRabbit committed review: 0 findings across all 3 changed files
- hosted CodeRabbit approval is attached to this exact commit
- unresolved review threads: 0

Hosted CI run 30724221323 is still in progress. This PR will not merge until the exact-head Copilot review and every required hosted job, including package builds, are green.

adamgell:
@coderabbitai full review

PR #496, opened 2026-08-05
fix(parser): stop panicking on non-ASCII decimal digits (#413)

### LINKED ISSUE #413: CCM parser panics on non-ASCII decimal digits in a time tail
Found by the independent review of PR #408. **Pre-existing, not introduced by that PR**, but worth fixing because this crate is published on crates.io and documented as pure with no I/O, so a panic on untrusted input is a real availability problem for anyone embedding it.

The `regex` crate's `\d` is Unicode-aware by default, so the time-tail capture can hold multi-byte digits while the hour, minute, and second groups remain ASCII and parse successfully. The byte-indexed slices that follow then panic on a character boundary.

Reproduced at three sites in `crates/cmtraceopen-parser/src/parser/ccm.rs`:
- `truncate_subsecond_to_millis`
- `split_ccm_time_tail`
- `split_legacy_public_time_tail`

Crashing inputs include a tail of `.١٢٣` (three Arabic-Indic digits, six bytes), and the mixed forms `.123١`, `.12١٢`, and `.1234١`. The reviewer confirmed the identical panic at the pre-#408 parent, so #408 did not introduce it.

The reason to fix it now rather than later: #408 rewrote one of the three sites, and its frozen table test claims to pin "every shape of CCM fractional-second tail". That claim holds only for ASCII shapes. Either the guard should exist or the claim should be narrowed, and the guard is clearly the better answer.

Suggested fix: a `value.is_ascii()` guard before the byte-indexed slicing, plus one row in the frozen ambiguity table covering a non-ASCII digit tail so the gap cannot reopen.

Refs #406, #317.

adamgell:
## Summary
- Restrict critical numeric regex captures to ASCII `[0-9]` so Unicode Nd digits cannot enter byte-indexed slicing paths.
- Guard CCM/simple/timestamped/reporting_events fractional helpers with ASCII-digit checks before `value[..n]` slices.
- Make DNS wire-label decoding char-boundary safe (label length + unexpected-character advance).
- Add unit + public-surface regression tests for the panic class documented in #413.

## Why
`regex`'s `\d` matches `\p{Nd}`. Multi-byte digits (e.g. Arabic-Indic `١٢٣`) made:
- `truncate_subsecond_to_millis` / time-tail splits in CCM
- fractional millis / timezone parsing
- reporting-events detection timestamps
- simple-format ms truncation
- DNS label extraction

panic with `byte index N is not a char boundary`. This crate is pure and published; panics on untrusted logs are an availability bug.

## Test plan
- [x] `cargo test --lib` in `crates/cmtraceopen-parser` (656 passed)
- [x] `cargo test --test issue_413_non_ascii_digits`
- [x] `cargo clippy --all-targets -- -D warnings` (clean)
- [x] Frozen CCM time-tail ambiguity table still green

Fixes #413

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log parsing stability for timestamps, fractional seconds, and DNS-style data containing non-ASCII characters.
  * Prevented crashes when malformed or Unicode numeric input is encountered.
  * Invalid timestamp and wire-format values are now rejected safely instead of causing parsing failures.

* **Tests**
  * Added regression coverage for Unicode digits, multibyte characters, malformed lengths, and invalid fractional timestamps.

### DIFF
--- a/crates/cmtraceopen-parser/src/parser/ccm.rs
+++ b/crates/cmtraceopen-parser/src/parser/ccm.rs
@@ -28,12 +28,14 @@ fn ccm_re() -> &'static Regex {
     CELL.get_or_init(|| {
         Regex::new(concat!(
             r#"<!\[LOG\[(?P<msg>[\s\S]*?)\]LOG\]!>"#,
-            r#"<time="(?P<h>\d{1,2}):(?P<m>\d{1,2}):(?P<s>\d{1,2})\.(?P<time_tail>\d+(?:[+-]\d+)?)""#,
-            r#"\s+date="(?P<mon>\d{1,2})-(?P<day>\d{1,2})-(?P<yr>\d{4})""#,
+            // ASCII digits only: regex \d matches Unicode Nd and feeds byte-indexed
+            // slices that panic on multi-byte digits (issue #413).
+            r#"<time="(?P<h>[0-9]{1,2}):(?P<m>[0-9]{1,2}):(?P<s>[0-9]{1,2})\.(?P<time_tail>[0-9]+(?:[+-][0-9]+)?)""#,
+            r#"\s+date="(?P<mon>[0-9]{1,2})-(?P<day>[0-9]{1,2})-(?P<yr>[0-9]{4})""#,
             r#"\s+component="(?P<comp>[^"]*)""#,
             r#"\s+context="(?P<context>[^"]*)""#,
-            r#"\s+type="(?P<typ>\d)?""#,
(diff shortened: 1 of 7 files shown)


adamgell:
Superseded by the independently reviewed and merged PR #502. Issue #413 is fixed on main at 5cc7da2004691eeea9f3f606404f121818eb8c50 with the broader CCM and CmtLog Unicode structural-digit coverage. Closing this older head unmerged to keep the queue truthful.

PR #501, opened 2026-08-05
fix(deps): raise chrono floor to 0.4.35

### LINKED ISSUE #417: cmtraceopen-parser declares chrono 0.4 but uses an API added in 0.4.35
Flagged by CodeRabbit on PR #407 and verified.

`crates/cmtraceopen-parser/Cargo.toml` declares:

```toml
chrono = { version = "0.4", features = ["serde"] }
```

but PR #407 introduces `chrono::DateTime::<chrono::Utc>::from_timestamp_millis` at `crates/cmtraceopen-parser/src/sccm/client/deployment.rs:1521`. That associated function was added in **chrono 0.4.35**.

This builds fine here because `Cargo.lock` resolves to 0.4.45. It is still wrong, and it matters more than usual for this crate specifically:

1. **cmtraceopen-parser is published on crates.io.** A downstream consumer whose lockfile already pins any 0.4.0 through 0.4.34 satisfies the declared `"0.4"` requirement and will fail to compile. Their build breaks on our declaration, not on their choice.
2. It breaks under `-Z minimal-versions`, which is the usual way this class of defect is caught in CI.

The same call also exists at `src-tauri/src/sysmon/evtx_parser.rs:501`, where `src-tauri/Cargo.toml` makes the identical `"0.4"` declaration. That one is not published, so the consequence is limited to a minimal-versions build, but the declaration is equally inaccurate.

**Fix:** raise both declarations to `chrono = { version = "0.4.35", features = ["serde"] }`. That is the minimum that actually compiles the code as written, and it is a permissive requirement (any 0.4.x at or above 0.4.35 satisfies it), so it costs downstream consumers nothing beyond accuracy.

**Worth considering separately:** adding a minimal-versions check to CI would catch this class automatically. It is exactly the kind of defect that is invisible locally because the lockfile hides it, and this repository has now had two defects in one night whose common feature was being invisible to the checks that were running.

Refs #322, #317.

adamgell:
## Summary
- raise the published `cmtraceopen-parser` chrono requirement to `0.4.35`
- raise the Tauri app chrono requirement to the same accurate floor

## Verification
- `cargo check -p cmtrace-open`
- `npx tsc --noEmit`
- temporary floor-resolution check: `cargo update -p chrono --precise 0.4.35 && cargo check -p cmtraceopen-parser -p cmtrace-open`

Closes #417

## Summary by CodeRabbit

* **Chores**
  * Updated the date and time handling library to a newer compatible release.
  * Retained support for serializing date and time values.

### DIFF
--- a/crates/cmtraceopen-parser/Cargo.toml
+++ b/crates/cmtraceopen-parser/Cargo.toml
@@ -21,7 +21,7 @@ categories = [
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 regex = "1"
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4.35", features = ["serde"] }
 encoding_rs = "0.8"
 log = "0.4"
 thiserror = "2"
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -49,7 +49,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 base64 = "0.23"
 regex = "1"
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4.35", features = ["serde"] }
 encoding_rs = "0.8"
 log = "0.4"
 thiserror = "2"


adamgell:
Independent critic verdict: **ACCEPT** at frozen SHA `e9ad3ea032181cac8ef988e80468d6846f9f5ec5`.

- Both and only direct Chrono declarations resolve as `^0.4.35`.
- A disposable detached copy was forced to exact `chrono 0.4.35`; `cmtraceopen-parser --all-targets` and `cmtrace-open` both compiled.
- Parser also compiled for `x86_64-pc-windows-msvc` at the exact floor.
- `Cargo.lock` correctly remains unchanged in this PR.
- No compatibility layer, fallback, or unrelated dependency change.
- `git diff --check` passed.

Residual: a full local Windows Tauri cross-check is unavailable on macOS without MSVC headers; hosted Windows CI is the authoritative remaining gate.

Issue #410, opened 2026-08-01
Log viewer displays wrong milliseconds and a fabricated timezone offset for signless CCM timestamps


adamgell:
Found while fixing #406. This one is on the **public surface users actually look at**, and it is currently pinned in a test as intended behavior.

`split_legacy_public_time_tail` in `crates/cmtraceopen-parser/src/parser/ccm.rs` hands the **final digit** of any signless fractional tail to the timezone offset and re-truncates what remains:

| tail | viewer displays | viewer offset | reality |
|---|---|---|---|
| `.123` | `.012` | +3 min | 123 ms, no offset |
| `.123456` | `.123` | +6 min | 123456 us, no offset |
| `.1234567` | `.123` | +7 min | IME 7-digit precision, no offset |
| `.1` | entry demoted to Plain, no timestamp at all | | |

So an ordinary CCM line `&lt;time="10:00:00.123"&gt;` with no explicit offset renders as `10:00:00.012` and its epoch timestamp is shifted three minutes earlier. That also perturbs the timeline and multi-file merge ordering, since both order on the epoch value.

This is pinned as the baseline at `crates/cmtraceopen-parser/tests/sccm_spine_contract.rs:4353`, `public_ccm_digit_only_timestamp_tails_match_pre_spine_baseline`, whose comment explains the intent: preserve the pre-spine public `LogEntry` behavior while SCCM provenance interprets the same tail independently. The compatibility goal was reasonable; the behavior it preserves is wrong.

Unlike the six-digit case fixed in #406, there is no reading of the documented `%03u%d` grammar under which the final digit of a three-digit tail is an offset. A fabricated 0 to 9 minute offset benefits nothing downstream.

**Related defect in the same path:** `truncate_subsecond_to_millis` (~line 165) parses fractions shorter than three digits as-is, so `.12` becomes 12 ms rather than 120 ms. It is unreachable in the corrected evidence path but is exactly why `.123` displays as `.012` today, so the two should be fixed together.

**Suggested fix, if the product call is to correct it:** point the public projection at the corrected `split_ccm_time_tail` and delete the `public_compatible` demotion path. Every observable change is a correction: displayed milliseconds become right for three-digit tails, `timezoneOffset` becomes 0 or absent instead of 1 through 9 for signless tails, timestamps move by under ten minutes, and single-digit-tail lines start parsing as CCM instead of Plain. The only cost is rewriting the deliberately pinned baseline test.

[...]

PR #502, opened 2026-08-05
fix(parser): reject Unicode decimal fields

### LINKED ISSUE #413: CCM parser panics on non-ASCII decimal digits in a time tail

adamgell:
## Summary
- restrict numeric log grammar captures to ASCII digits
- keep DNS wire-format parsing on UTF-8 character boundaries
- cover public parser, ReportingEvents detection, DNS, and Intune event tracking surfaces

## Verification
- `cargo test --test issue_413_unicode_panics`
- `rustfmt --check` on changed Rust files

Closes #413

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log parsing with consistent ASCII digit handling across timestamps, error codes, retries, inventory, and event fields.
  * Prevented crashes and infinite loops when processing malformed Unicode text in DNS names and numeric log fields.
  * Improved handling of invalid reporting-event data.

* **Tests**
  * Added regression coverage for Unicode digits, malformed fields, Unicode DNS names, and Intune event tracking.


adamgell:
Independent critic verdict at `d89a1dc694c023d35c10cd14b5364a44ffd926fe`: **REWORK**.

Blocking omission: `parser/cmtlog.rs` still uses Unicode-capable `\d` for structural numeric CCM fields and defaults failed conversions. A public CmtLog input with `time="14:30:00.١٢+000"` remains structured and fabricates `.000` instead of failing closed/preserving malformed input.

The existing CCM/simple/timestamped/reporting/DNS/event-tracker fixes otherwise passed the class audit. Rework is assigned to:

- restrict CmtLog structural numeric grammar to ASCII;
- add a public semantic regression for the reproduced input;
- strengthen changed-parser regressions to assert output/fallback and valid Unicode payload preservation, not only “did not panic.”

#410/#414 greedy signless-tail semantics remain separate.

adamgell:
Independent critic re-review: **ACCEPT** at frozen SHA `5bd464d3cdb462c267dc516cf43f422eae3779fd`.

- CmtLog Unicode structural digits now fail closed: raw line preserved, parse error recorded, no fabricated timestamp/component.
- Valid Unicode payload remains structured across CmtLog, CCM, Simple, timestamped, ReportingEvents, and Intune.
- DNS byte-length behavior is UTF-8 boundary-safe.
- Full parser: 1,846 passed.
- Strict parser Clippy, scoped formatting, and diff checks passed.
- Independent public-surface harness passed.
- #410/#414 signless-tail grouping/greediness is unchanged and was separately reproduced.

Residual risk is limited to that already-filed legacy signless-tail behavior, outside #413. Hosted CI is the remaining gate.

adamgell:
Final integration critic: **ACCEPT** at frozen merge SHA `bdad76b9bfe626819becc0f5628c98668ee5018b`.

- Exact parents: accepted #413 `5bd464d3...` and current main `8113e178...`.
- Both parents are ancestors; remerge diff has no resolution delta.
- #413 source/tests are byte-identical to the accepted parent.
- Relative to main, exactly the eight expected #413 paths remain.
- Both Chrono manifests retain `0.4.35`.
- Focused 7/7; full parser 1,846; strict Clippy; scoped formatting/diff all passed.

Residual integration risk is low. Restarted hosted CI is the final merge gate.

PR #503, opened 2026-08-05
fix: redact bare unspecified IPv6 in macOS exports

### LINKED ISSUE #416: IPv6 redaction rule leaked addresses and destroyed C++ symbol names in two files
Found while addressing CodeRabbit feedback on PR #402. **Fixed in that PR** (commit `d131aef7`), filed for the record because it was a live privacy leak, it was present in two independent files, and the second half of the defect was not reported by any reviewer.

## The rule expressed the inverse of its intent

The IPv6 redaction pattern used `\b` word-boundary anchors. A word boundary requires a word character on exactly one side, and `:` is not a word character, so the anchors did the opposite of what was intended in both directions.

**Direction one, addresses leaked.** Every `::`-compressed address either begins or ends with a colon and so had nothing to anchor against. These were exported verbatim: `::1`, `[::1]`, `fe80::`, a bare `::`, `peer ::1 connected`, and the `::ffff:` head of an IPv4-mapped address. Only forms that begin and end on a hex digit, such as `[fe80::1]:8080`, redacted correctly.

**Direction two, and this one no reviewer reported: the rule destroyed evidence.** A word boundary *does* exist between an identifier character and a colon, so C++ scope resolution matched:

```
std::__1::basic_string  ->  std[redacted:address]__1[redacted:address]basic_string
std::cout               ->  std[redacted:address]cout
```

The rule was corrupting backtrace symbol names, which are precisely the evidence a Console capture exists to carry. This is why simply widening the pattern would have been wrong: the fix had to close the leak without reopening the corruption.

## Present in two files

The identical pattern was live in both `crates/cmtraceopen-parser/src/intune/portal/ios_ipados/company_portal/diagnostics/redaction.rs` and `crates/cmtraceopen-parser/src/esp/redaction.rs`. The ESP copy's only IPv6 test used a middle-`::` form that begins and ends on hex, so it exercised neither failure mode. Both are fixed with regression tests.

## Class audit result

Every other `\b` anchor in both files was audited: `url`, `bearer`, `jwt`, `keyed_secret`, `tenant_id`, `device_id`, `guid`, `certificate_thumbprint`, `ipv4`, `mac`, and `app_identifier` all anchor against character classes that begin and end with word characters, so their boundaries are well defined. `email` deliberately carries no `\b`. **IPv6 was the only affected rule.**

The fix checks boundaries against neighbouring characters directly, mirroring the existing `redact_app_identifiers` scan, because the `regex` crate has no lookaround. A trailing `:` deliberately does not bind, since IPv4-mapped addresses arrive as `::ffff:`.

Over-redaction is covered by its own test: `std::__1::basic_string`, `std::cout`, `3:2`, `key:value`, and `10:00:00` all survive untouched.

**Lesson worth keeping:** a redaction test that only exercises the well-formed middle case can pass while the rule is broken at both edges. Redaction tests should cover the boundary forms specifically, in both directions: what must be redacted and what must survive.

Refs #372.

adamgell:
## Summary
- redact the bare unspecified IPv6 address `::` in macOS Company Portal direct-log and unified-log exports
- use Unicode-aware identifier boundaries so C++ symbols, underscore continuations, and non-ASCII identifier-adjacent text remain intact
- preserve delimiters between adjacent bare, full, compressed, and IPv4-mapped addresses
- cover both public export surfaces, including timestamps, MAC categories, ratios, `key:value`, versions, and GUID categories

## Validation
- `cargo test -p cmtraceopen-parser --test company_portal_macos_logs`
- `cargo test -p cmtraceopen-parser --test company_portal_macos_unified_log`
- `rustfmt --edition 2021 --check` on the four changed Rust files
- `git diff --check origin/main...HEAD`

Closes #416

## Summary by CodeRabbit

* **Bug Fixes**
  * IPv6 redaction now handles bare, compressed, bracketed, peer-list, and IPv4-mapped addresses in macOS Company Portal and unified logs.
  * Improved boundary detection prevents identifiers, timestamps, versions, GUIDs, MAC addresses, ratios, and C++ scope syntax from being misidentified as IPv6 addresses.
  * Redacted addresses embedded in supported text are handled without altering surrounding content.

* **Tests**
  * Added regression coverage for IPv6 formats, token boundaries, delimiters, and unrelated text preservation.

### DIFF
--- a/crates/cmtraceopen-parser/src/intune/portal/macos/company_portal/logs/redaction.rs
+++ b/crates/cmtraceopen-parser/src/intune/portal/macos/company_portal/logs/redaction.rs
@@ -104,31 +104,15 @@ fn mac_address_re() -> &'static Regex {
     })
 }
 
-/// Fully-expanded eight-group IPv6 or any `::`-compressed form, in group 1.
+/// Fully-expanded eight-group IPv6 or any `::`-compressed form.
 ///
-/// The address is **not** wrapped in `\b`. `\b` is a word-boundary assertion and
-/// `:` is not a word character, so it cannot anchor an address that begins or
-/// ends with a colon, which every `::`-compressed form does at one end or both.
-/// With `\b` on both sides `fd12:3456:789a::` matched nothing at all and exported
-/// verbatim, and `::1` was missed the same way.
-///
-/// This engine has no look-around, so the left edge is a *consumed* guard group
-/// and the right edge has no assertion at all. That asymmetry is deliberate:
-///
-/// * Consuming only the left delimiter keeps two addresses separated by a single
-///   character both matchable. A guard on both sides would eat the delimiter that
-///   the following address needs, and `^` cannot stand in for it because it
-///   anchors to the start of the text, not to where scanning resumed.
-/// * No right-edge assertion is needed because the alternatives are already
-///   self-limiting: they consume only hex and colons, and each requires either
-///   eight groups or a `::` run. A MAC (`00:11:22:33:44:55`) and a house-grammar
-///   timestamp (`08:18:00:104`) have neither, so neither can be claimed.
-///
+/// Deliberately unanchored: [`Redactor::replace_ipv6_addresses`] checks both
+/// Unicode-aware token boundaries without consuming either delimiter.
 fn ipv6_address_re() -> &'static Regex {
     static CELL: OnceLock<Regex> = OnceLock::new();
     CELL.get_or_init(|| {
         Regex::new(
-            r"(?i)(?:^|[^0-9A-Za-z:])((?:[0-9a-f]{1,4}:){7}[0-9a-f]{1,4}|(?:[0-9a-f]{1,4}:)+:(?:[0-9a-f]{1,4}(?::[0-9a-f]{1,4})*)?|::(?:[0-9a-f]{1,4}(?::[0-9a-f]{1,4})*)?)",
+            r"(?i)(?:(?:[0-9a-f]{1,4}:){7}[0-9a-f]{1,4}|(?:[0-9a-f]{1,4}:)+:(?:[0-9a-f]{1,4}(?::[0-9a-f]{1,4})*)?|::(?:[0-9a-f]{1,4}(?::[0-9a-f]{1,4})*)?)",
         )
         .expect("ipv6 address pattern must compile")
     })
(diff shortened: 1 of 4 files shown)


adamgell:
Independent privacy critic verdict at `10b51498894178dbc57d5e3e60eb420d93c44fd1`: **REWORK**.

Blocking over-redaction: both new macOS IPv6 regexes permit underscore as a left delimiter, so public export corrupts `std::__1::basic_string` by matching the `1::ba` suffix.

Observed outputs include:

- direct: `std::__[redacted:ip:001]sic_string`
- unified: `std::__[redacted:host]sic_string`

Bare-address, adjacent-address, mapped-address, timestamp, ratio, key/value, MAC, and GUID checks otherwise passed. Rework is assigned to align identifier boundaries with the safe lanes—including underscore/non-ASCII adjacency—and expand both public-export matrices before another critic round.

adamgell:
Independent critic re-review: **ACCEPT** at frozen SHA `cf2ba2275127fcd5f958218254384443d5cbb2e9`.

- `std::__1::basic_string` and `std::process::exit` remain unchanged on both public export surfaces.
- Bare `::` redacts at start/middle/end/brackets.
- Adjacent bare/full/compressed addresses retain delimiters.
- IPv4-mapped forms are fully redacted.
- Unicode/underscore-bound lookalikes, timestamps, ratios, key/value text, and versions survive.
- MAC/GUID retain their intended categories.
- Rejected identifier-bound matches allocate no direct-log placeholders; repeated valid values deterministically reuse tokens.
- Focused suites 27/27 and 38/38, full parser/doc tests, strict Clippy, scoped formatting, and diff checks passed.

Residual: combining-mark identifier adjacency is outside the explicit boundary definition; no current reproduction. Hosted CI remains the gate.

adamgell:
Final integration critic: **ACCEPT** at frozen merge SHA `cdebf70eee54920ea39c4e2738e0bea94ea0e23f`.

- Exact parents: accepted #416 `cf2ba227...` and current main `8113e178...`.
- Both parents are ancestors; remerge diff is empty.
- Relative to main, only the accepted two macOS redactors and two public-export tests remain.
- Those four files are byte-identical to the accepted parent.
- Both Chrono manifests retain `0.4.35`; lock resolves `0.4.45`.
- Focused 27/27 and 38/38, full parser/doc suite, strict Clippy, scoped formatting/diff all passed.

No new residual risk beyond the accepted boundary-policy limits. Restarted hosted CI is the final merge gate.

PR #504, opened 2026-08-05
fix(parser): treat signless CCM tails as fractions

### LINKED ISSUE #410: Log viewer displays wrong milliseconds and a fabricated timezone offset for signless CCM timestamps

### LINKED ISSUE #414: cmtlog.rs still carries the original greedy timestamp regex
Found by the independent review of PR #408, which fixed this defect family in `parser/ccm.rs`.

`crates/cmtraceopen-parser/src/parser/cmtlog.rs` (around line 40) still contains the exact greedy pattern that was the *first* wrong implementation of the CCM time tail:

```
(?P<ms>\d+)(?P<tz>[+-]*\d+)
```

Because `\d+` is greedy, it takes all but one digit and hands the **final digit** to the timezone offset. For a tail of `.123` that yields ms=12 and tz=+3 minutes.

This does not reach the SCCM evidence spine, since `scan_logical_records` is CCM-only, so it is not the same severity as #406. But it is a public `LogEntry` producer, which means it is the same class of user-visible defect described in #410, in a second file.

Worth deciding together with #410, since both are about whether the public projection should keep reproducing a known-wrong timestamp for backward compatibility. If #410 is fixed, this should be fixed in the same change; if #410 is deliberately kept, this should at least carry a comment saying so, so the next reader does not treat it as an oversight.

Refs #410, #406, #317.

adamgell:
Closes #410
Closes #414

- Treat every unsigned CCM/CmtLog tail digit as fractional seconds.
- Scale one/two fractional digits to milliseconds and truncate longer fractions.
- Retain only explicit signed timezone offsets and fail closed for signless chronology.
- Preserve #413 ASCII structural digit rejection and Unicode payload support.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CCM and CmtLog timestamp handling for fractional seconds and explicit timezone offsets.
  * Preserved millisecond-level precision while rejecting malformed or ambiguous timestamp formats.
  * Prevented incorrect timezone inference from unsigned timestamp tails.
  * Improved Unicode content handling without treating Unicode digits as timestamp structure.
  * Updated inventory recovery validation and timestamp ordering for more reliable results.

* **Documentation**
  * Added reference documentation for updated timestamp behavior and validation rules.


adamgell:
Independent critic verdict at `94aeea5151539557ff40737b532b8090149b37fc`: **REWORK — functional behavior accepted; committed verification plan is not executable/accurate yet**.

Required corrections:

1. Split the invalid Cargo command that supplies two positional test filters.
2. Replace nonexistent `timeline_does_not_order_signless_fraction_as_utc` with implemented `signless_ccm_fraction_is_not_timeline_orderable`; the current command exits successfully while running zero intended tests.
3. Replace package-wide formatting/“all commands exit 0” claims with scoped `rustfmt --check`, because unrelated baseline formatting drift is known.
4. Add direct CmtLog `+240` coverage alongside the accepted `-240`/`+000` paths.
5. Clean remaining test-local “signless-offset” terminology where the contract now asserts fraction-only rejection.

Parser behavior itself passed the public projection, provenance, chronology, CmtLog, #413, inventory, full parser, and strict Clippy audit.

adamgell:
Independent critic re-review: **ACCEPT** at frozen SHA `45c77a7bd6a9aa8bce1961071946c44217b11cce`.

- Every documented Cargo command is syntactically valid and names existing tests.
- Scoped Rustfmt covers every changed Rust file; no false package-wide formatting claim remains.
- Direct CmtLog `.12+240` with trailing `color` parses as offset `Some(240)` and display `.120`; `-240` and Unicode structural rejection remain covered.
- Corrected public fraction, signed offset, no-signless-ordering, inventory rejection, and client-management mutation tests all passed.
- #413 regression 7/7, full parser, strict Clippy, scoped format, and diff checks passed.

Residual risk is low. Fresh hosted CI is the final gate.

adamgell:
Hosted Windows diagnostics failed one unrelated app test at frozen SHA `45c77a7bd6a9aa8bce1961071946c44217b11cce`:

`watcher::tail::tests::test_tail_reader_bounds_unterminated_inventory_continuation`

Observed: entry count was correct (1), but `parse_errors` was 0 instead of 1. The test uses the InventoryAdaptor tail path; PR #504 changes CCM/CmtLog timestamp parsing and scoped SCCM fixtures, not watcher/tail or inventory parsing. The identical test passed on combined main run 30987714977. Full Rust PR job passed.

Failed-job rerun and an independent regression/flake audit are active. No merge and no speculative code change until that evidence returns.

PR #505, opened 2026-08-05
test: tighten deployment fixture expectations

### LINKED ISSUE #411: Deployment corpus overstates validated families and declares an unused selectionState
Two corpus defects found by the independent review of PR #407, both in the merged deployment corpus under `crates/cmtraceopen-parser/tests/fixtures/sccm/client/deployment/`.

**1. `rotation-boundary` declares a family validated that never yielded a record.**

That scenario's `expected.json` declares `extractionProfile.validatedArtifactFamilies: ["client-app-intent", "client-content"]`, but neither rotation fragment in the `client-content` family ever formed a complete logical record. The #322 reducer reports only `["client-app-intent"]`, which is correct.

This matters beyond a bookkeeping mismatch. It is exactly the failure class this program exists to eliminate: **a coverage claim that survives when its source does not.** A corpus that asserts a family was validated when nothing was read from it would, if a reducer matched it, teach that reducer to claim validation it cannot support. The reviewer's judgment was that the reducer is strictly better than the corpus here and that the corpus should be corrected rather than the deviation tolerated indefinitely.

**2. `extractionProfile.selectionState` is declared in all twelve scenarios and consumed by nothing.**

Every `expected.json` carries `selectionState: "selected"`. `grep` for `selectionState` or `selection_state` across `crates/cmtraceopen-parser/src/` returns nothing, and `SccmDeploymentExtractionProfile` has no such field. Either the field means something and a reducer should be required to produce it, or it is vestigial and should be removed so "the scenarios reproduce exactly" is true without an asterisk.

Both are corpus-side decisions rather than reducer bugs, which is why they are filed separately from the PR #407 review.

Refs #322, #317.

adamgell:
## Summary
- remove unconsumed `extractionProfile.selectionState` from all twelve deployment expectations
- reject unowned extraction-profile expectation fields in the corpus harness
- retain the rotation-boundary claim for only `client-app-intent`, because its content fragments do not produce a logical record

## Verification
- `cargo test -p cmtraceopen-parser --test sccm_client_deployment`
- `rustfmt --check crates/cmtraceopen-parser/tests/sccm_client_deployment.rs`
- `jq -e .` across all deployment expected JSON fixtures

Closes #411

## Summary by CodeRabbit

* **Tests**
  * Updated deployment test fixtures to remove obsolete extraction profile selection-state data.
  * Added validation ensuring extraction profiles contain only the fields required by deployment tests.
  * Continued validating profile identifiers, version requirements, key types, and artifact families.

### DIFF
--- a/crates/cmtraceopen-parser/tests/fixtures/sccm/client/deployment/bits-transfer-failure/expected.json
+++ b/crates/cmtraceopen-parser/tests/fixtures/sccm/client/deployment/bits-transfer-failure/expected.json
@@ -4,7 +4,7 @@
   "scenario":"bits-transfer-failure",
   "stateChain":["intent","requirements","locateContent","transfer","cache","enforce","detect","report"],
   "analysisContract":{"independentReducer":true,"consumesPolicyReducerOutput":false,"policyCoverageRequired":false,"crossSideCorrelationPerformed":false},
-  "extractionProfile":{"selectionState":"selected","profileId":"deployment-client-5.00.test-v1","sourceVersionPrefix":"5.00.TEST.","contentVersionRequired":true,"keyKinds":["assignmentId","bitsJobId","ciId","contentId","contentVersion","distributionPointHostHandle","packageId","requestId"],"validatedArtifactFamilies":["client-app-intent","client-content"]},
+  "extractionProfile":{"profileId":"deployment-client-5.00.test-v1","sourceVersionPrefix":"5.00.TEST.","contentVersionRequired":true,"keyKinds":["assignmentId","bitsJobId","ciId","contentId","contentVersion","distributionPointHostHandle","packageId","requestId"],"validatedArtifactFamilies":["client-app-intent","client-content"]},
   "reorderedInputDeterministic":true,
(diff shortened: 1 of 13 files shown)


adamgell:
Independent review verdict: **REWORK / DO NOT MERGE** at `5207585debc7e899e30c1d7115564876a2093bd5`.

The issue premise is stale:

- `SccmDeploymentProfileSelectionState` is a production enum.
- `SccmDeploymentExtractionProfile.selection_state` serializes publicly as `selectionState`.
- The reducer produces it from evidence.
- SCCM correlation consumes it to distinguish validated from unknown authority.
- Selected and unselected branches have explicit correlation tests.

Removing the twelve fixture fields and their direct assertion would delete a real contract. The new exact-key harness is therefore incorrect because it bans the sixth owned field.

The rotation-boundary defect was already fixed before this PR by `eb1b8ba50fee8fb7a6630f8d29b951c3ba88d0c4`; the PR base already contains only `client-app-intent`. This PR adds no rotation correction.

Closing unmerged.

PR #508, opened 2026-08-05
fix(sccm): remove fixture identity allowlists from intake

### LINKED ISSUE #409: intake.rs hardcodes test-fixture string literals in production source
Found while analyzing the intake-versus-corpus schema question. This is a defect independent of that decision and blocks every option equally.

When a manifest sets `syntheticFixture: true` (every corpus does), `crates/cmtraceopen-parser/src/sccm/server/windows/intake.rs` validates identifiers against **hardcoded literal allowlists of its own eleven fixtures**:

- `safe_manifest_artifact_id` (~line 918): roughly 19 literal artifact ids
- `safe_lineage_id` (~962): 17 literals
- `safe_path_fingerprint` (~988): 13 literals
- `safe_optional_handle` (~1021): 4 literals
- `normalize_topology` (~311): requires `siteCode == "LAB"` and `captureHost` in {`LAB-CM01`, `LAB-MP01`}
- `safe_source_id` (~948): an **unconditional** 8-value allowlist, not gated on the synthetic flag at all

Consequences:

1. **Adding any fixture requires editing production source.** A test corpus cannot grow without a change to `src/`, which inverts the normal dependency direction and means fixture authorship is gated on a parser review.
2. **It silently caps what intake can ever accept.** 47 of 159 server artifacts across `hierarchy_and_replication`, `provider_and_admin_service`, and a few others have source ids absent from `safe_source_id`, so they are rejected regardless of how well-formed they are.
3. **The synthetic gate does not fully protect production behavior**, because `safe_source_id` applies unconditionally.

The allowlists are clearly there for a good reason: keeping synthetic fixtures from smuggling identity-bearing values into a privacy-sensitive contract. The problem is the mechanism, not the intent. Options worth weighing: move the literals into a test-only module or a fixture-declared allowlist file; derive the permitted set from the catalog rather than restating it; or replace the literal list with a grammar (the same closed-grammar approach lanes #326 and #319 converged on for their public surfaces) so new conforming values pass without a source edit.

Refs #317.

adamgell:
## Summary
- replace server-intake synthetic fixture identity lists with bounded field grammars
- make the server source catalog the sole owner of declared source IDs
- preserve opaque production handles and fail-closed malformed/identity-bearing rejection

## Verification
- `cargo test -p cmtraceopen-parser --test sccm_server_intake`
- `CARGO_INCREMENTAL=0 cargo test -p cmtraceopen-parser`
- `CARGO_INCREMENTAL=0 cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings`
- scoped `rustfmt --check` and `git diff --check origin/main`

Closes #409

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SCCM server-intake validation for conforming synthetic artifact, host, subject, site-code, path, and lineage identities.
  - Preserved validation of production source IDs, topology relationships, source versions, and rotation limits.
  - Improved rejection of malformed, unsafe, and inconsistent identity formats.
  - Synthetic WSUS artifacts are accepted when structurally valid, while catalog-bound fields remain protected.
  - Standardized content-addressed identity handling across server roles and workflows.

- **Documentation**
  - Added guidance for server-intake identity validation updates.


adamgell:
Final independent integration verdict: ACCEPT at exact SHA `f40f552c7dde5c697497cc5140a122e866003d05`. Parents are accepted #409 `2af3c9a` plus current main/#506 `162f1c2`; reconstructed integration tree matches exactly, with no manual product delta. SCCM intake 66/66 and Device Inventory 24/24 pass; both library routes occur exactly once and resolve; changed Rust formatting and diff checks pass. Exact-head merge is authorized after hosted CI and all package jobs are green.

PR #521, opened 2026-08-08
fix(clipboard): register Info pane copies in Windows clipboard history

### LINKED ISSUE #520: Bug with Windows clipboard history
If i select a log line in the main pane and copy it with any of the right-click options (Copy Message, etc), the text gets copied into clipboard history as expected.
But copying from the Info pane at the bottom behaves weirdly:
* If i select some text in the info pane (part of the message) and use "right-click + copy", i can still paste that selection but it isn't visible in the clipboard history
* Another possibly related bug is that after selecting text in the Info pane and using Ctrl+C the whole line gets copied into clipboard (clipboard history) rather than just the selection. Not sure how it's supposed to work but when a keyboard shortcut is displayed with the menu option i'd expect consistency. Ideally Ctrl+C should copy just the selection in the Info pane if there's any, and the whole line only when nothing is selected.

Image

This is version 1.4.0

P.S. Great app, even though takes a bit longer to start compared to the original CMTrace.

adamgell:
## Summary

Fixes both symptoms reported in #520:

1. **Info pane right-click Copy never appeared in Windows clipboard history (Win+V).** The Info pane has no custom context menu, so its Copy is the WebView2 native copy, which the OS clipboard accepts but clipboard history does not capture. All plugin-routed copies (log list context menu, keyboard fallback) do register, which is exactly the asymmetry the reporter observed.
2. **Ctrl+C with a selection in the Info pane copied the whole log line.** The global Ctrl+C fallback in `use-keyboard.ts` never consulted `window.getSelection()`.

## Changes

- `src/hooks/use-keyboard.ts`: the Ctrl+C fallback now returns without preventing default when a non-empty text selection exists, so the native copy takes the selection. With nothing selected, the whole-entry copy is unchanged.
- `src/hooks/use-clipboard-history-mirror.ts` (new): document-level `copy` listener that re-writes natively copied text through the Tauri clipboard plugin so it registers in clipboard history. Windows-only; skips events an app handler already `preventDefault`ed; handles both DOM selections and input/textarea ranges.
- `src/components/layout/AppShell.tsx`: mounts the mirror next to `useKeyboard()`.

This closes the class, not just the instance: every native copy path (Info pane menu, Ctrl+C on selections, inputs) is mirrored, and every app-initiated copy already went through the plugin.

## Not done / assumptions

- Not verified on a live Windows machine: the claim that plugin writes reach history and native WebView2 writes do not rests on the reporter's own observations plus known WebView2 clipboard quirks. A manual Win+V check on Windows before merge would confirm.
- The mirror writes plain text only. On Windows the native copy is plain text in this app, so nothing is downgraded.

## Verification

- `npx vitest run`: 657 tests, 52 files, all passing (8 new tests: selection priority, whitespace-only selection, input-range mirroring, defaultPrevented guard, no-selection no-op, platform gating)
- `npx tsc --noEmit`: clean

Closes #520

## Summary by CodeRabbit

* **New Features**
  * Clipboard copies from selected text are now mirrored to Windows clipboard history.
  * Native text selection in inputs, textareas, and the page is supported.
  * Whole-entry copying remains available when no text is selected.

* **Bug Fixes**
  * Prevented events, empty selections, and non-Windows platforms avoid unintended clipboard changes.
  * Keyboard copying now preserves native behavior when text is actively selected.

### DIFF
--- a/src/components/layout/AppShell.tsx
+++ b/src/components/layout/AppShell.tsx
@@ -37,6 +37,7 @@ import { useFileWatcher } from "../../hooks/use-file-watcher";
 import { useIntuneAnalysisProgress } from "../../workspaces/intune/use-intune-analysis-progress";
 import { useSysmonAnalysisProgress } from "../../workspaces/sysmon/use-sysmon-analysis-progress";
 import { useKeyboard } from "../../hooks/use-keyboard";
+import { useClipboardHistoryMirror } from "../../hooks/use-clipboard-history-mirror";
 import { useDragDrop } from "../../hooks/use-drag-drop";
 import { useFileAssociation } from "../../hooks/use-file-association";
 import { useFileAssociationPrompt } from "../../hooks/use-file-association-prompt";
@@ -292,6 +293,9 @@ export function AppShell() {
   useIntuneAnalysisProgress();
   useSysmonAnalysisProgress();
   useKeyboard();
+  // Re-write native WebView copies through the clipboard plugin so they
+  // register in Windows clipboard history (#520)
+  useClipboardHistoryMirror();
   useDragDrop();
   // Handle file path passed via OS file association at startup
   useFileAssociation();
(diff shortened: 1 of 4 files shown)


adamgell:
@coderabbitai full review

adamgell:
bug testing is done. ready to push

PR #460, opened 2026-08-03
feat(intune): parse Company Portal Windows LocalState logs

adamgell:
## Scope

Implements the first raw-format parser for Windows Company Portal `LocalState`
`Log_<n>.log` files for #366 (part of #356). Restacked on current `main` via
non-destructive merge (no force-push).

It adds content-confirmed detection, logical-record framing, typed
`LogEntry`-compatible projection, version-scoped raw evidence documents,
synthetic/sanitized fixtures, default redaction, and the required parser/UI
registration. The pure parser crate performs no filesystem collection or live
Windows access.

## Evidence and safety boundaries

- A Company Portal path or generic `Log_<n>.log` name only nominates a
  candidate; record structure must confirm it.
- The one validated grammar profile is explicitly version-scoped. Unknown app
  versions are experimental/low-confidence coverage, not validated facts.
- Evidence is redacted by default; the local unredacted projection is explicit.
- Rotation, malformed/truncated records, encoding variants, unrelated UWP
  logs, generic timestamped logs, and same-time distinct activities are all
  covered by synthetic fixtures.
- Empty input is not reported as successful (`Available`) coverage.
- Aggregate/live tail framing preserves Company Portal logical continuations
  with bounded amendments and physical-line provenance.
- This PR does **not** claim live Windows collection/acceptance, a second
  real-version capture, or a semantic root-cause engine.

## Restack note (lane A)

Merged `origin/main` (`8064b5aa`) into `codex/intune-366-review-fixes-r119`
to clear `CONFLICTING` state. Sole content conflict was `.gitattributes`;
resolved by retaining byte-sensitive fixture rules for parser corpora
(`-text -whitespace`) including Company Portal and SCCM paths.

## Local verification after restack

```text
cargo test --locked -p cmtraceopen-parser --test company_portal_windows_logs
  -> 34 passed
cargo test --locked -p cmtrace-open --test parser_supported_formats
  -> 32 passed
cargo test --locked -p cmtrace-open --lib watcher::tail
  -> 29 passed
npm test -- --run src/stores/log-store.test.ts src/hooks/use-file-watcher.test.tsx src/lib/tail-payload-validation.test.ts
  -> 98 passed
cargo clippy --locked -p cmtraceopen-parser --all-targets -- -D warnings
  -> clean
cargo check --locked -p cmtraceopen-parser --target wasm32-unknown-unknown
  -> clean
npx tsc --noEmit
  -> clean
git diff --check origin/main...HEAD
  -> clean
```

Refs #366

[...]

PR #522, opened 2026-08-08
ci(probe): widened ESP Windows job on plain main (do not merge)

adamgell:
Probe for the PR #460 `STATUS_ENTRYPOINT_NOT_FOUND` failure: main + only the widened ESP Windows steps from #460. If the ESP job here fails the same way, the widened job itself is the trigger and #460's code is exonerated. Will be closed after one CI cycle.


adamgell:
Probe complete: reproduced the four sccm_client_management_fixture_contract path-separator failures on plain main + widened job, confirming the latent Windows bug lives on main (fix rides PR #460 as commit 46e7a6a1). The app_lib loader question was answered separately by a dumpbin diagnostic on the #460 branch; the target-dir isolation fix is on that branch.

PR #540, opened 2026-08-09
feat(collection): capture LocaleMetaData so exported channels are self-describing

adamgell:
Closes part of #539 (Phase 2 groundwork).

## Why now

`wevtutil export-log` emits only the binary records. Rendering an event description then requires the originating provider to be registered on whatever machine opens the file, which is never true for an analyst workstation and is **impossible on macOS or Linux**.

The timing argument is the important one: **metadata cannot be recovered after the fact**. A bundle collected today without LocaleMetaData is permanently unresolvable once that machine is rebuilt. Every bundle collected between now and whenever the reader ships is lossy for no reason. This lands ahead of the parser deliberately.

## What changed

After each successful export the collector runs `wevtutil archive-log`, producing a `LocaleMetaData\<name>_<lcid>.MTA` sidecar carrying the provider's message strings. Each sidecar is recorded in the manifest as an `event-log-metadata` artifact with its own hash.

- Archive failures degrade to an explicit **coverage gap**, not silence. The evtx stays `collected`.
- Locale is left to the collecting machine's default so captured strings match what the operator saw on that endpoint. The emitted LCID is read from the produced filename rather than predicted.
- `-SkipLocaleMetadata` opts out.

Cost measured on the lab host: **+241 KB** on a zipped seven-channel bundle (411,890 to 653,155 bytes).

## Two pre-existing bugs fixed along the way

The collector **could not run at all** against the shipped profile, on either host. Both were found while trying to verify the change end to end, and both are fixed here because the feature could otherwise never execute:

1. `ConvertFrom-Json -Depth 20` at the profile load. `-Depth` does not exist on `ConvertFrom-Json` in Windows PowerShell 5.1 and fails parameter binding. That is the host the collector is deployed under.
2. The optional-array check read values through `Get-ObjectPropertyValue`, which returns through the pipeline. That enumerates a single-element array such as `arguments: ["/status"]` down to a bare string, so the shipped profile was rejected as malformed.

The existing suite missed both because its fixture profile only used empty arrays, which enumerate to nothing and read as absent. New tests cover single-element, empty, and scalar cases, plus the shipped profile itself.

Worth noting on the second fix: my first attempt changed `Get-ObjectPropertyValue` to return with a unary comma. That is wrong. Callers collect it with `@(...)`, which treats a comma-returned array as a single pipeline item and nests it one level deep. The tests caught it. The fix now reads the raw property at the one site that needs to distinguish array from scalar, matching what `Assert-ProfileRequiredArray` already does.

## Verification

Run end to end on a Windows 11 x64 lab host (`RING0IVY24-01`, build 26200) under **both** hosts:

| Host | Result | evtx | MTA | manifest records |
|---|---|---|---|---|
| Windows PowerShell 5.1 | completed 83.6 s | 9 | 9 | 9 |
| pwsh 7.6 | completed 82.9 s | 9 | 9 | 9 |

Both previously failed at profile load. Artifact statuses `collected=68 missing=5`, the 5 being channels genuinely absent on that host and correctly recorded as coverage gaps.

Pester 14/14. Script parses clean, ASCII-only per the PS 5.1 rule.

## Not included

Parsing `.MTA` is not in this PR. The container format is decoded and specced in [#539](https://github.com/adamgell/cmtraceopen/issues/539#issuecomment-5231738369) (magic `MTAFile\0`, three sections EVT/MSG/PUB); the inner section layouts still need reverse engineering. This PR only ensures the data is captured so that work has something to read.

## Summary by CodeRabbit

* **New Features**
  * Event-log evidence collection now archives channel locale metadata alongside exported logs.
  * Collection manifests record collected, missing, and failed metadata artifacts, including locale identifiers.
  * Added an option to skip locale metadata collection.
* **Bug Fixes**
  * Improved reporting of missing, unreadable, and failed metadata artifacts.
  * Enhanced compatibility with Windows PowerShell 5.1.
* **Tests**
  * Expanded coverage for metadata handling, profile validation, and optional data.

### DIFF
--- a/references/collection/Invoke-CmtraceEvidenceCollection.ps1
+++ b/references/collection/Invoke-CmtraceEvidenceCollection.ps1
@@ -1,4 +1,4 @@
-﻿[CmdletBinding()]
+[CmdletBinding()]
 param(
     [string]$OutputRoot = (Join-Path $env:ProgramData 'CmtraceOpen\Evidence'),
     [string]$BundleLabel = 'intune-endpoint-evidence',
--- a/references/collection/Invoke-CmtraceEvidenceCollection.ps1
+++ b/references/collection/Invoke-CmtraceEvidenceCollection.ps1
@@ -1,4 +1,4 @@
-[CmdletBinding()]
+﻿[CmdletBinding()]
 param(
     [string]$OutputRoot = (Join-Path $env:ProgramData 'CmtraceOpen\Evidence'),
     [string]$BundleLabel = 'intune-endpoint-evidence',
@@ -824,19 +824,22 @@ function Export-EventChannelLocaleMetadata {
         return $records
     }
 
+    # Test-Path is inside the try deliberately. $ErrorActionPreference is 'Stop' for this script, so
+    # a provider or I/O fault while probing the folder would otherwise abort the entire collection
+    # rather than recording a failed artifact for this one channel.
     $metadataFiles = @()
-    if (Test-Path -LiteralPath $metadataFolder) {
-        try {
+    try {
+        if (Test-Path -LiteralPath $metadataFolder -ErrorAction Stop) {
             $metadataFiles = @(Get-ChildItem -LiteralPath $metadataFolder -Filter ('{0}_*.MTA' -f $baseName) -File -ErrorAction Stop)
         }
-        catch {
-            # An access or I/O fault here is a failure, not an absence. Reporting it as 'missing'
-            # would claim the sidecar was never produced when it may simply be unreadable.
-            $notes = 'Could not enumerate {0}: {1}' -f $metadataFolder, (Protect-SecretText -Text $_.Exception.Message)
-            $records.Add((New-ArtifactRecord -Category 'event-log-metadata' -Family $Family -RelativePath $unresolvedRelativePath -OriginPath $Channel -Status 'failed' -ParseHints @('mta') -Notes $notes))
-            Add-ObservedGap -ObservedGaps $ObservedGaps -Status 'failed' -Origin $Channel -Reason $notes
-            return $records
-        }
+    }
+    catch {
+        # An access or I/O fault here is a failure, not an absence. Reporting it as 'missing' would
+        # claim the sidecar was never produced when it may simply be unreadable.
+        $notes = 'Could not enumerate {0}: {1}' -f $metadataFolder, (Protect-SecretText -Text $_.Exception.Message)
+        $records.Add((New-ArtifactRecord -Category 'event-log-metadata' -Family $Family -RelativePath $unresolvedRelativePath -OriginPath $Channel -Status 'failed' -ParseHints @('mta') -Notes $notes))
+        Add-ObservedGap -ObservedGaps $ObservedGaps -Status 'failed' -Origin $Channel -Reason $notes
+        return $records
     }
 
     if ($metadataFiles.Count -eq 0) {
--- a/references/collection/Invoke-CmtraceEvidenceCollection.ps1
+++ b/references/collection/Invoke-CmtraceEvidenceCollection.ps1
@@ -139,6 +139,37 @@ function Get-LocaleMetadataRelativePath {
     return (Join-RelativePath -Left $normalizedPath.Substring(0, $lastSeparatorIndex) -Right 'LocaleMetaData')
 }
 
+function Get-LocaleMetadataLcid {
+    <#
+        .SYNOPSIS
+        Extracts the locale identifier wevtutil.exe encoded into an .MTA sidecar's file name.
+
+        .DESCRIPTION
+        Sidecars are named <exported-log>_<lcid>.MTA, for example device-management-admin_1033.MTA.
+        The LCID is read back from the name rather than predicted, because the collecting machine's
+        default locale decides it and the exported log's own base name may itself contain
+        underscores.
+    #>
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$MetadataFileName
+    )
+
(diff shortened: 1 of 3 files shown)


adamgell:
Both open findings addressed.

**BOM (new).** Real: both copies opened with `EF BB BF` before `[CmdletBinding()]`. Nothing in either file is non-ASCII, so the BOM bought nothing and cost the one thing it can — any path reading the script as bytes rather than through PowerShell's own encoding detection sees three junk characters ahead of the first token. Piping the file into a remote shell is exactly such a path, and it is how this collector gets deployed. Fixed in `b0af5922`.

Verified on the real thing rather than reasoned about, on PowerShell 5.1.26100:

```
BOM present    : False
First 3 bytes  : 5B 43 6D
PARSE OK       : 7303 tokens
SCRIPTBLOCK OK : 55124 chars
```

Checked the rest of the class too: neither copy holds any other non-ASCII byte, so that was the only instance. Both copies stay byte-identical, which the parser crate asserts.

**Test-Path inside the try (18:13Z).** Already fixed on this branch. `Test-Path` is inside the `try` with `-ErrorAction Stop`, with a comment recording why, and the catch still produces the failed artifact and the observed gap:

```powershell
try {
    if (Test-Path -LiteralPath $metadataFolder -ErrorAction Stop) {
        $metadataFiles = @(Get-ChildItem -LiteralPath $metadataFolder -Filter ... -ErrorAction Stop)
    }
}
catch {
    # An access or I/O fault here is a failure, not an absence.
```

adamgell:
@coderabbitai full review

Both open findings are addressed (BOM stripped in b0af5922; the Test-Path case was already inside the try with -ErrorAction Stop). Please re-review the current head.

adamgell:
All three addressed in `096c0577`.

**wevtutil.exe unguarded (real).** `$ErrorActionPreference` is `Stop`, so the command being absent from PATH or failing to launch took down the entire collection instead of recording one failed channel. This is the same class as the `Test-Path` case fixed earlier on this branch — that fix closed one instance and left the sibling open. Losing every other artifact in a bundle over one sidecar is the worst available outcome, so it now records a failed `event-log-metadata` record plus the observed gap and returns.

**`$profile` shadowing (real).** Renamed to `$testProfile`. Audited the class rather than the instance: no other PowerShell automatic variable is shadowed anywhere in either collector copy or the test suite.

**Source-text tests (real, and the most useful of the three).** Five tests asserted the collector's text matched regexes, which passes just as happily when the behaviour is wrong and breaks on a harmless reformat. They now load `Export-EventChannelLocaleMetadata` through the suite's existing AST loader and invoke it against a stubbed `wevtutil.exe` and a temp sidecar folder, asserting returned record status, `relativePath`, notes carrying the parsed LCID, `hashes.sha256`, and the matching `$ObservedGaps` entries. Function definitions win over external commands in PowerShell's resolution order, so the stub is reached without touching the collector — which also means these run off Windows.

I checked the new tests can actually fail, rather than trusting a green run:

| mutation | result |
|---|---|
| remove the `wevtutil` guard | 1 failed |
| report an enumeration fault as `missing` instead of `failed` | 1 failed |
| unmutated | 26 passed |

Only the `SkipLocaleMetadata` assertion stays text-based: it is a parameter on the script itself, so there is no unit to invoke.

Verified on the real runtime too — PowerShell 5.1.26100 parses the collector to 7,375 tokens with no errors and builds a scriptblock from all 54,738 characters. Both copies stay byte-identical, ASCII, no BOM.

adamgell:
## Charter review — PR #540

**Reviewed head:** `096c05778404ae0010065c78f29b724b6c0aaeaf`

### Findings (most severe first)

**No blocking findings.** The implementation keeps the two collector copies byte-identical, records collected/missing/failed LocaleMetaData outcomes as coverage states, records LCIDs from the emitted filename, hashes collected sidecars, and uses a deterministic file-shaped placeholder for unresolved outcomes. The opt-out path skips metadata capture without changing the event-log artifact.

### Named gate states

- **CI checks: PASS.** Live PR checks at the reviewed head are successful: Rust check/test, Rust MSRV (Ubuntu and Windows), TypeScript, E2E, ESP Diagnostics (Windows), CodeQL analyses, and macOS/Windows/Linux builds. `git diff --check` is clean.
- **CodeRabbit review state at head (`approved_at_head`): PASS.** The live CodeRabbit review is `APPROVED` on `096c0577`; the CodeRabbit status check is also `pass`. Earlier actionable findings were verified against the final code and addressed.
- **Hermes charter review: POSTED, no open blocking findings.** Focused verification passed: PowerShell AST parse (7,375 tokens) and Pester 6.0.0: **26 passed, 0 failed, 0 skipped**.
- **Contract-layer conformance: PASS / NOT APPLICABLE as a reducer contract gate.** This PR changes the Windows evidence-collection script and tests only; it does not touch a reducer, parser crate, identity/correlation, chronology, terminal precedence, or cross-lane semantic contract. Applicable evidence-contract checks pass: failures and absences remain explicit manifest/gap states, rather than being reported as collected evidence.

### Review feedback rejected

**None.** No review feedback was rejected as invalid. CodeRabbit findings concerning LCID notes, invocation/enumeration failure handling, the BOM, and behavior-based tests were confirmed valid and are addressed. Copilot's suppressed concern about directory-shaped unresolved paths was also valid and is addressed by `*_unknown-lcid.MTA`; it was not rejected.

### Gate state

**Merge gate: READY on the reviewed head, subject to Adam's merge action.** Hermes did not approve or merge the PR.

This review covered the `origin/main...origin/feat/event-log-locale-metadata` diff, both collector copies, the Pester suite, live PR checks/reviews, and the applicable collection/evidence contracts; it did not cover unrelated repository code, live Windows endpoint collection, or production tenant/device data.

PR #542, opened 2026-08-09
fix(scripts): let the Windows prereq installer run at all on PowerShell 5.1

adamgell:
## The documented Windows setup command could never have worked

From `CLAUDE.md` and the README:

```
powershell -ExecutionPolicy Bypass -File .\scripts\Install-CMTraceOpenBuildPrereqs.ps1
```

It fails immediately, before installing anything:

```
Split-Path : Cannot bind argument to parameter 'Path' because it is an empty string.
At ...\Install-CMTraceOpenBuildPrereqs.ps1:10 char:45
+     [string]$RepoRoot = (Split-Path -Parent $PSScriptRoot)
```

## Root cause, isolated rather than guessed

`$RepoRoot` defaulted to `(Split-Path -Parent $PSScriptRoot)`. **Windows PowerShell 5.1 leaves `$PSScriptRoot` empty while evaluating parameter defaults on an *advanced* script**, and this script is advanced because of its `[CmdletBinding(SupportsShouldProcess = $true)]` attribute.

Confirmed on Windows 11 build 26200 with three minimal scripts differing only in that attribute:

| Script | `$PSScriptRoot` in param default |
|---|---|
| plain `param()` | populated, works |
| **+ `[CmdletBinding(SupportsShouldProcess=$true)]`** | **empty, throws** |
| + CmdletBinding + `Set-StrictMode` | empty, throws |

`pwsh` 7 populates it either way, which is why this survived: anyone testing under PowerShell 7 sees it work. 5.1 is the default shell on Windows and the one the documented command invokes.

## Fix

`RepoRoot` is resolved in the script body, where `$PSScriptRoot` is populated. An explicitly passed `-RepoRoot` still wins.

## Class, not instance

Checked every `.ps1` in the repository for the same shape. This is the only one that reads `$PSScriptRoot` inside a `param()` default, so the class is closed rather than just this occurrence.

## How it was found

While installing the toolchain on the lab host in order to verify the event log work in #539 against real Windows. The installer is confirmed to progress past this point and install Node, VS Build Tools, and WebView2 once the argument is supplied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed prerequisite installation script failures in Windows PowerShell 5.1.
  * The repository path is now resolved reliably at runtime when no path is provided.

### DIFF
--- a/scripts/Install-CMTraceOpenBuildPrereqs.ps1
+++ b/scripts/Install-CMTraceOpenBuildPrereqs.ps1
@@ -7,11 +7,20 @@ param(
     [switch]$InstallRepoDependencies,
     [switch]$SkipValidation,
 
-    [string]$RepoRoot = (Split-Path -Parent $PSScriptRoot)
+    [string]$RepoRoot
 )
 
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
+
+# Resolved here rather than as a param default. Windows PowerShell 5.1 leaves $PSScriptRoot empty
+# while evaluating defaults on an advanced script, which this is because of the CmdletBinding
+# attribute above, so the default threw before the script could run at all. Verified on 5.1: the
+# same default works without CmdletBinding and fails with it. pwsh 7 populates it either way, which
+# is why this survived.
+if ([string]::IsNullOrWhiteSpace($RepoRoot)) {
+    $RepoRoot = Split-Path -Parent $PSScriptRoot
+}
 
 function Write-Step {
     param(


adamgell:
## Hermes charter gate-state report

### Findings (most severe first)

No blocking, warning, or informational correctness findings survived verification.

The changed path is `scripts/Install-CMTraceOpenBuildPrereqs.ps1`. The branch removes the `$PSScriptRoot`-based parameter default and resolves `$RepoRoot` in the script body when the argument is null, empty, or whitespace. This preserves an explicitly supplied `-RepoRoot` and avoids the Windows PowerShell 5.1 advanced-script parameter-default failure described by the PR.

Independent checks performed:

- The exact `origin/main...origin/fix/prereqs-psscriptroot` diff is one file and is limited to this change.
- `git diff --check` passes.
- The changed PowerShell parses successfully with `pwsh`.
- The repository-wide search found no second instance of `$PSScriptRoot` in a `param()` default. Other uses are in script bodies or tests.
- The current head is `4c1f90304808f2eec74fc9472af857d271df6b26`, and GitHub reports the PR as mergeable with a clean merge state.

### Named gate states

- **CI checks: PASS.** All 17 reported checks pass, including Rust check/test, TypeScript check, Linux/Windows/macOS builds, E2E, ESP Diagnostics, MSRV, CodeQL, and the CodeRabbit status check.
- **CodeRabbit review at head (`approved_at_head`): PASS.** CodeRabbit submitted `APPROVED` for head `4c1f90304808f2eec74fc9472af857d271df6b26`.
- **Contract-layer conformance: PASS / not applicable.** This is an installer-script-only change and does not touch reducer or evidence-lane code, public parser contracts, chronology/correlation semantics, coverage semantics, or redaction behavior. The reducer contract checklist and ADR-specific conformance questions therefore have no changed surface to evaluate.
- **Hermes charter review: PASS.** This report is the required Hermes charter review; it has no open blocking findings.

### Review feedback explicitly rejected

- **CodeRabbit’s BOM comment is rejected as non-blocking and not a defect.** The BOM is present at byte 0 of the script and is the standard compatibility marker for UTF-8 scripts consumed by Windows PowerShell 5.1. The script otherwise contains only ASCII content. Removing it would not improve the reviewed fix and could make UTF-8 decoding less reliable under the exact legacy host this PR targets. The comment is outside the changed diff and does not affect the runtime-root fix.
- The Copilot review supplied only a restatement of the change and no actionable finding; there is no defect to adopt from it.

### Closing scope

This review covered the exact PR diff against `main`, the installer’s PowerShell parsing and root-resolution paths, repository-wide duplicate-shape search, CI state, CodeRabbit’s review state at head, and the applicable contract gate; it did not claim live Windows PowerShell 5.1 execution because that environment is not available in this review session.

PR #543, opened 2026-08-10
refactor(intune): thin Framework v1 extraction (mapping, test support, invariant docs)

adamgell:
Thin PR 4 of the Reducer Framework. An inspection cycle concluded that almost nothing should be extracted; this implements exactly the three items that survived that conclusion. No reducer behavior changes.

## 1. Extracted: the `IntuneAccessState` <-> `IntuneArtifactStatus` mapping

Three copies of a pure 7-arm total bijection existed. **Every arm was verified identical before replacing** (see "arm-by-arm verification" below). Both directions now live in `crates/cmtraceopen-parser/src/intune/evidence.rs`, beside the two enums:

- `artifact_status_for_access_state(&IntuneAccessState) -> IntuneArtifactStatus`
- `access_state_for_artifact_status(&IntuneArtifactStatus) -> IntuneAccessState`

Both are exhaustive `match` with no `_` arm, so a new variant on either enum is a compile error rather than a silently defaulted status in whichever lane copied the table last. Signatures take `&` because neither enum is `Copy` and adding `Copy` would be an unrelated public API change.

The exhaustiveness test moved from `win32/sources.rs` to `evidence.rs` and now also pins the inverse direction (the Store lane's half), which nothing tested before.

Lane updates:

| Lane | Before | After |
|---|---|---|
| `apps/windows/win32/sources.rs` | local `coverage_status` | calls `artifact_status_for_access_state` |
| `apps/windows/microsoft_store/reducer.rs` | local `access_state_for` | calls `access_state_for_artifact_status` |
| `enrollment/windows/autopilot/sources.rs` | `declared_status` and `access_state`, both 7-arm tables over the lane-local `AutopilotCaptureState` | `access_state` kept (it is this lane's own collector vocabulary); `declared_status` is now `artifact_status_for_access_state(&self.access_state())` |

Autopilot keeps `access_state` because its source enum is `AutopilotCaptureState`, which is not shared and should not be. Only the second hop, `IntuneAccessState -> IntuneArtifactStatus`, was ever the shared part.

### Arm-by-arm verification

| `IntuneAccessState` | `IntuneArtifactStatus` | win32 `coverage_status` | store `access_state_for` (inverse) | autopilot `declared_status` / `access_state` |
|---|---|---|---|---|
| `Available` | `Available` | match | match | `Captured` -> both, consistent |
| `Missing` | `Missing` | match | match | `Absent` -> both, consistent |
[...]

Issue #364, opened 2026-07-31
Intune Windows device: model compliance evaluation and reporting evidence


adamgell:
## Parent

Part of #356.

## Outcome

Implement `cmtraceopen_parser::intune::device::windows::compliance` to explain the local evaluation and reporting path for Intune device compliance without equating Conditional Access denial, stale cloud state, or missing user evaluation with a failed local setting.

## Input contract

- normalized MDM/device-management events from validated Windows channels;
- exported MDM diagnostics report sections relevant to compliance and enrolled configuration sources;
- optional IME custom-compliance/HealthScripts evidence, explicitly labeled as that workload;
- optional imported Intune device/per-setting compliance facts with capture time and provenance;
- optional sign-in/Conditional Access result facts as supplemental downstream evidence only;
- Windows build, enrollment, active-user/device context, and policy metadata when supplied.

Source contracts must identify which facts are device-local, user-scoped, service-reported, custom-compliance script output, or downstream access decisions.

## Stable identity/privacy

Use policy/setting ID, CSP/resource path, evaluation ID, enrollment/device key, user/device scope, custom-compliance policy/run ID, and report ID when explicit.

UPN, user SID, tenant/device IDs, custom script output, and access-resource details are sensitive. Redact by default. Preserve enough stable synthetic identity to show whether evidence belongs to the same local evaluation.

## Output/state contract

Model: policy known/received, local prerequisite met/unmet, setting evaluated compliant/noncompliant/error/not-applicable, custom-compliance detection outcome, aggregate device result, report queued/submitted/failed/stale, service result observed, and insufficient evidence.

Required distinctions:

- local device compliant but service state stale;
- local setting noncompliant;
- evaluation error/unsupported;
- user-targeted policy not evaluated in the current context;
- custom-compliance script failure versus a noncompliant discovery result;
- downstream Conditional Access denial with and without a matching compliance state;
- missing expected local/service sources.

Do not diagnose Conditional Access policy itself in this module. A sign-in fact can show downstream impact only when explicit matching identity and time provenance exist.

## Implementation/file boundary

Create modules under `crates/cmtraceopen-parser/src/intune/device/windows/compliance/` and reuse reviewed normalized Windows event/report inputs from the configuration issue rather than duplicating them.

Add:

- `crates/cmtraceopen-parser/tests/intune_windows_compliance.rs`
- `crates/cmtraceopen-parser/tests/fixtures/intune/windows/compliance/<scenario>/`.

[...]

PR #545, opened 2026-08-10
fix(intune): make the Compliance reducer independent of caller order

adamgell:
Stacked on #543 (`framework/pr4-thin-extraction`), which rewired `tests/intune_windows_compliance.rs` to the shared helpers this PR also touches.

The Compliance reducer pushed observations in raw caller order and never sorted them before folding. Six merges read vector position instead of record content. One of them **flipped an exported conclusion**, which is what makes this more than cosmetic.

## RED first

Every test below was written and run before any production change.

### 1. Conclusion-flipping: `last_submission_at` and the staleness gate (severe)

`reducer.rs` set `phase.last_submission_at = submitted_at` unconditionally on every `Submitted` row, so the value was whichever submission the caller listed **last**. That field feeds the staleness gate, so two submissions straddling `latest_local` exported `Submitted` or `Stale` purely on vector order.

`two_submissions_straddling_the_evaluation_do_not_flip_the_report_state`:

```
assertion `left == right` failed: the staleness gate must not read whichever submission the caller listed last
  left: Submitted
 right: Stale
```

### 2. Lane-level permutation over the scenario corpus

`permuting_sibling_records_does_not_change_the_analysis`, in the style of Configuration's, using the shared `support::wire` from #543 for byte-for-byte comparison.

**The corpus alone could not fail this test.** All 16 scenarios are single-record-per-setting bundles: none carries two observations sharing a grouping token, two report submissions, or two rows describing one policy. Sweeping only the corpus would have shipped a gate that passes vacuously. The test therefore sweeps the corpus **plus** `contended_bundle()`, a synthetic input in which every order-sensitive merge is actually contended. The pinned fixture matrix is left untouched.

With the contended case, pre-fix:

```
assertion `left == right` failed: contended-bundle: input order must not change the analysis
```

The two snapshots differed in: `reporting.state` `stale` vs `submitted`, `reporting.lastSubmissionAt` `09:00` vs `11:00`, `reporting.reportId` `alpha-report` vs `zebra-report`, `settings[0].displayName` `Alpha…` vs `Zebra…`, `settings[0].error` `0x80070002` vs `0x80070005`, `settings[0].scope` `user` vs `device`, `policies[0].state` `notReceived` vs `received`, `policies[0].scope` `user` vs `device`, and the `compliance-local-evaluation-error` finding summary quoting a different setting name.

### 3. Duplication and 4. irrelevant evidence

`re_collecting_the_same_records_does_not_change_the_analysis` and `irrelevant_evidence_does_not_change_an_existing_conclusion` both passed pre-fix and are kept as the regression net the lane did not have. They are honest passes, not RED — recorded as such rather than dressed up.

### 5. Additional RED found by inspection during the work

Two more members of the same class, both confirmed RED before fixing:

`a_queued_row_and_an_unknown_row_do_not_swap_the_report_state` — the `Submitted`/`Failed` arms of the report fold were rank-monotone, but the remaining three tokens fell through to a plain last-writer-wins:

```
assertion `left == right` failed: the weaker report tokens must fold by rank like the stronger ones do
  left: Unknown
 right: Queued
```

Plus `two_submission_rows_do_not_let_caller_order_pick_the_report_id_or_error` (`Some("zebra")` vs `Some("alpha")`), `a_merged_setting_display_name_does_not_follow_caller_order` (`Some("Zebra setting")` vs `Some("Alpha setting")`), `a_merged_setting_error_does_not_follow_caller_order` (`0x80070005` vs `0x80070002`), `two_records_disagreeing_about_scope_do_not_follow_caller_order` (`Device` vs `User`), `two_policy_rows_do_not_let_caller_order_pick_the_policy_state` (`Received` vs `NotReceived`).

## Semantic decision per field changed

| Field | Decision | Reasoning |
|---|---|---|
| `reporting.last_submission_at` | **Latest submission by timestamp**; a single submission is reported whatever its zone; several are ordered only when *every* one carries a known zone, otherwise `None` | The gate asks "does the service hold a copy of the most recent evaluation?", which depends on the latest *successful* submission, never on vector position. One zone-less stamp among several means we cannot say which is last; naming one would let the gate declare `Stale` on a guess. `None` leaves the state at `Submitted` — the claim the evidence does support |
| `reporting.state` | Fold by explicit rank `Submitted > Failed > Stale > Queued > Unknown` | Preserves the documented precedence and extends it to the three tokens that previously fell through to last-writer-wins. Rank is injective over the five variants, so there are no ties to break |
| `reporting.report_id` | Smallest stated | Content-determined where "first" was the collector's serializer. Every submission row stays cited |
| `reporting.error` | Smallest by `(decimal, hex, raw)` | Configuration's `terminal_error` ordering |
| `settings[].scope` | **Agreed value, else `Unknown`** (conservative) | Scope decides whether a setting reads as a user-targeted policy that was never evaluated (missing coverage) or as a device failure. Genuinely ambiguous with equal authority, so ADR-003 says report neither |
| `settings[].display_name` | Smallest stated | Follows Configuration's `smallest_stated`. Not dropped on disagreement: the findings quote this label, and falling back to the grouping token loses real usefulness for a cosmetic field. Both records stay cited |
| `settings[].error` | Smallest by `(decimal, hex, raw)` | Configuration's `terminal_error`, verbatim |
| `settings[].key.*` | Smallest stated; `identity_contradiction` recomputed as "the stating records do not agree" | Same answer as the old running comparison, without reading order |
| `settings[].evaluated_at` | Orderable beats unorderable, newest wins, ties settled on smallest content | Preserves existing semantics exactly; only the tie and all-unorderable cases changed, which previously resolved to whichever was listed last. Deliberately does **not** refuse like `last_submission_at`: this is provenance, and `latest_evaluation` re-filters to orderable instants before drawing any conclusion |
| `policies[].state` / `.scope` | **Agreed value, else `Unknown`** (conservative) | Two sources saying received and not received are equally authoritative and there is no third fact to break the tie |
| `prerequisites[].detail` | Smallest stated among rows carrying the winning state | The state fold was already a rank; the detail followed the *last* unmet row |
| `custom_compliance`, `service_results`, `access_impact.decisions` | Exported arrays sorted by evidence ref | Sets of results, not sequences. Nothing reads their position, but the caller's order was reaching the serialized snapshot |

Left unchanged as already order-independent, per the brief: `settings[].state` (rank-monotone), the definite/indefinite setting-state fold (disagreement → `Contradictory`), `fold_freshness`, and the access-linkage correlation.

`merge_setting` also canonicalizes its observations up front. That changes no output today, since every merge in it is individually content-determined; the comment says so rather than overstating it. It is there so a field added later that walks the vector inherits a stable order.

## Fixtures

**No fixture expectation changed.** `every_scenario_matches_its_expected_reduction` passes unmodified across all 16 scenarios, and no `expected.json` was touched. The corpus never exercised a contended merge, which is exactly why it did not catch this.

## Gates

| Gate | Result |
|---|---|
| `cargo test -p cmtraceopen-parser` | **2452 passed, 0 failed** (baseline 2442 + 10 new: 7 reducer unit tests, 3 lane tests) |
| `cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings` | clean |
| `cargo check --workspace` | clean |
| `cargo check -p cmtraceopen-parser --target wasm32-unknown-unknown` | clean |
| `git diff --check` | clean |

No `tauri::test::mock_app()`. No rustfmt run on files not created here.

## Out of scope, for a separate PR

The lane's fixture corpus has no scenario in which two records describe the same setting, policy, or submission. That gap is why the order dependence survived review and why the permutation gate needed a synthetic bundle to have teeth. A corpus scenario covering contended evidence would be worth adding, but it changes the pinned `SCENARIOS` matrix from issue #364 and belongs in its own PR.

Assessability, confidence, linkage, and redaction were not touched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Compliance results are now consistent regardless of record processing order.
  - Duplicate records no longer change reported outcomes.
  - Conflicting policy, prerequisite, setting, report, service, and access data now resolves predictably.
  - Statuses, timestamps, errors, evidence, identifiers, and displayed details use stable selection and ordering.
  - Irrelevant or competing evidence is less likely to affect final compliance results.
  - Results remain reliable when timestamp information is incomplete or partially unknown.

### DIFF
--- a/crates/cmtraceopen-parser/src/intune/apps/windows/microsoft_store/reducer.rs
+++ b/crates/cmtraceopen-parser/src/intune/apps/windows/microsoft_store/reducer.rs
@@ -24,7 +24,7 @@
 //! somebody else's transaction.
 
 use crate::intune::evidence::{
-    IntuneAccessState, IntuneArtifactCoverage, IntuneArtifactStatus, IntuneErrorCode,
+    access_state_for_artifact_status, IntuneAccessState, IntuneArtifactCoverage, IntuneErrorCode,
     IntuneEvidenceRef, IntuneFindingConfidence, IntuneNamedValue, IntuneObservationContext,
     IntuneParseState, IntuneProvenance, IntuneSensitivity, IntuneTimestamp, IntuneTimestampKind,
 };
@@ -91,18 +91,6 @@ pub fn analyze_store_bundle(artifacts: &[StoreSourceArtifact]) -> StoreAnalysis 
 }
 
 // ── Observation extraction ──────────────────────────────────────────────────
-
-fn access_state_for(status: &IntuneArtifactStatus) -> IntuneAccessState {
-    match status {
-        IntuneArtifactStatus::Available => IntuneAccessState::Available,
-        IntuneArtifactStatus::Missing => IntuneAccessState::Missing,
-        IntuneArtifactStatus::PermissionDenied => IntuneAccessState::PermissionDenied,
-        IntuneArtifactStatus::Capped => IntuneAccessState::Capped,
-        IntuneArtifactStatus::Skipped => IntuneAccessState::Skipped,
-        IntuneArtifactStatus::ParseFailed => IntuneAccessState::Failed,
-        IntuneArtifactStatus::Unsupported => IntuneAccessState::Unsupported,
-    }
-}
 
 fn collect_from_artifact(artifact: &StoreSourceArtifact, observations: &mut Vec<StoreObservation>) {
     let Some(payload) = &artifact.payload else {
@@ -189,7 +177,7 @@ fn ime_context(
         // IME records quote package paths and app display names.
         sensitivity: IntuneSensitivity::Sensitive,
         parse_state,
-        access_state: access_state_for(&artifact.status),
+        access_state: access_state_for_artifact_status(&artifact.status),
     }
 }
 
@@ -1046,7 +1034,7 @@ fn build_coverage(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::intune::evidence::IntuneSourceKind;
+    use crate::intune::evidence::{IntuneArtifactStatus, IntuneSourceKind};
     use crate::intune::normalized::{NormalizedEventLevel, NormalizedWindowsEvent};
 
     fn context(artifact: &str, record: u64, kind: IntuneSourceKind) -> IntuneObservationContext {
--- a/crates/cmtraceopen-parser/src/intune/apps/windows/win32/sources.rs
+++ b/crates/cmtraceopen-parser/src/intune/apps/windows/win32/sources.rs
@@ -12,7 +12,7 @@
 //! would silently classify as an unknown artifact.
 
 use crate::intune::evidence::{
-    IntuneAccessState, IntuneArtifactCoverage, IntuneArtifactStatus, IntuneSourceKind,
+    artifact_status_for_access_state, IntuneAccessState, IntuneArtifactCoverage, IntuneSourceKind,
 };
 
 use super::models::Win32SourceKind;
@@ -224,22 +224,6 @@ pub fn classify_artifact(
     }
 }
 
-/// Map a caller-declared access state onto the shared coverage status.
-///
-/// Binding these two vocabularies in one place is what stops a bundle declaring
-/// an artifact absent and then reporting coverage for it as available.
-pub(super) fn coverage_status(access_state: &IntuneAccessState) -> IntuneArtifactStatus {
-    match access_state {
-        IntuneAccessState::Available => IntuneArtifactStatus::Available,
-        IntuneAccessState::Missing => IntuneArtifactStatus::Missing,
-        IntuneAccessState::PermissionDenied => IntuneArtifactStatus::PermissionDenied,
-        IntuneAccessState::Capped => IntuneArtifactStatus::Capped,
-        IntuneAccessState::Skipped => IntuneArtifactStatus::Skipped,
-        IntuneAccessState::Failed => IntuneArtifactStatus::ParseFailed,
-        IntuneAccessState::Unsupported => IntuneArtifactStatus::Unsupported,
-    }
-}
-
 /// The shared source kind that matches a Win32 artifact kind.
 pub(super) fn intune_source_kind(kind: Win32SourceKind) -> IntuneSourceKind {
     match kind {
@@ -273,7 +257,7 @@ pub(super) fn artifact_coverage(
     IntuneArtifactCoverage {
         artifact_id: input.artifact_id.clone(),
         family: coverage_family(kind).to_owned(),
-        status: coverage_status(&input.access_state),
+        status: artifact_status_for_access_state(&input.access_state),
         detail,
         observed_at_utc: input.captured_at_utc.clone().unwrap_or_default(),
         evidence: Vec::new(),
@@ -403,27 +387,7 @@ mod tests {
         assert_eq!(split_rotation("aé.lo"), ("aé.lo".to_string(), Some(0)));
     }
 
-    #[test]
-    fn every_access_state_maps_to_exactly_one_coverage_status() {
-        for (state, status) in [
-            (
-                IntuneAccessState::Available,
-                IntuneArtifactStatus::Available,
-            ),
-            (IntuneAccessState::Missing, IntuneArtifactStatus::Missing),
-            (
-                IntuneAccessState::PermissionDenied,
-                IntuneArtifactStatus::PermissionDenied,
-            ),
-            (IntuneAccessState::Capped, IntuneArtifactStatus::Capped),
-            (IntuneAccessState::Skipped, IntuneArtifactStatus::Skipped),
-            (IntuneAccessState::Failed, IntuneArtifactStatus::ParseFailed),
-            (
-                IntuneAccessState::Unsupported,
-                IntuneArtifactStatus::Unsupported,
-            ),
-        ] {
-            assert_eq!(coverage_status(&state), status);
-        }
-    }
+    // `every_access_state_maps_to_exactly_one_coverage_status` moved to
+    // `crate::intune::evidence` with the mapping it asserts, where it now also
+    // pins the inverse direction the Microsoft Store lane depends on.
 }
--- a/crates/cmtraceopen-parser/src/intune/device/windows/compliance/reducer.rs
+++ b/crates/cmtraceopen-parser/src/intune/device/windows/compliance/reducer.rs
@@ -628,6 +628,22 @@ fn latest_submission_at<'a>(
     }
 }
 
+/// Where a record sits on the timeline, or that it cannot be placed on one.
+///
+/// [`normalized_timestamp`] declines a stamp whose zone is unknown, so `None`
+/// here means "not placeable", never "earliest". The leading flag says so in the
+/// ordering: `false < true`, so every placeable record sorts before every
+/// unplaceable one and the dated run stays contiguous instead of being prefixed
+/// by a block whose position would read as "before everything".
+///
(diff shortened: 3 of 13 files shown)


adamgell:
Pushed fixes for all five findings as `56d38935..eb906832` (fast-forward, no force-push).

**Finding 1 was correct and was the blocking one.** All three comparators sorted on the `evidence` vector alone and `sort_by` is stable, so the PR left caller order reaching three exported arrays while claiming to remove it. Fixed RED-first: `e7a42014` adds the failing test, `56c03b81` makes the comparators total.

Reachability is proven rather than assumed: all three arrays set `evidence: vec![fact.context.evidence_ref.clone()]` (a one-element vector), and `sources::stage_record` reads `context.evidenceRef` verbatim from the collector envelope with no uniqueness constraint and no dedup anywhere downstream. A collector stamping one artifact-level ref across the rows it emits produces exactly this shape.

Chain order was chosen deliberately, not alphabetically: `custom_compliance` and `service_results` lead with containment (`policy_id` then the narrower id) so one policy's rows cannot interleave with another's, with time last; `decisions` leads with `occurred_at` because a denial is an event and that array reads as a sequence. Both timestamp links reuse the existing `timestamp_content_key` so a zone-less stamp still orders by content rather than collapsing to `None`.

**Findings 3, 4, 5 fixed as described.** Finding 5's new test is arranged so a wrong answer is visible in the export rather than only in the unnamed instant: the `Unspecified` submission predates the local evaluation and the zoned one postdates it, so a naive implementation flips `Submitted` to `Stale`. Mutation-verified by deleting the guard arm.

**Finding 2 was valid and reachable, not skipped.** `smallest_stated` took `Some("")` at face value while `is_identified`, `grouping_token`, and `contradicts` all already dropped blanks, so a group merged on its URI could export `settingId: ""` that `contradicts` had ruled out as no statement. Reachable through the report path only (`classify_setting_report` copies verbatim from JSON where `""` is valid); the event path cannot produce it because `named_any` trims and rejects blanks. Handled with the existing `usable_identifier` helper rather than a new one.

Verified independently before requesting re-review: RED reproduces at `e7a42014` (1 failed), full suite green at `eb906832`, no `expected.json` touched.

@coderabbitai full review

adamgell:
Round 3 pushed as `1355be13` (RED) and `1c5c198c` (fix). The finding was correct: the round-2 chains were still enumerated, so records differing only in an omitted field tied and kept caller order.

**Fixed as a class rather than by chaining four more fields**, because enumeration fails again the moment anyone adds a field. Each site keeps its named chain and appends `content_key` — the record's own `serde_json` serialization — as a final tie-breaker. Two records now tie exactly when they serialize identically, which is exactly when swapping them leaves the exported snapshot byte-identical; at that point caller order has nothing left to decide.

`derive(Ord)` was considered and rejected on two grounds. It does not work as-is (`IntuneTimestamp`, `IntuneErrorCode`, `IntuneNamedValue` derive only `PartialEq, Eq`), and making it work would put a semantic ordering on `IntuneTimestamp` whose derived form sorts by `raw_text` — deliberately a content key here, not a chronology. It would also make struct field declaration order the primary sort, discarding the round-2 reading intent and reordering exports whenever a field is moved.

**The class audit found two further sites in the same file**, which is why this is round 3 and not round 4:
- `named_data` in `merge_setting` compared `name` then `value` — the two fields `IntuneNamedValue` has *today*. That struct lives in the shared evidence module, so widening it there would have reintroduced caller order here, in a different module, with nothing to fail. It also broke `dedup`, which reads full equality.
- `observations` sorted on the evidence ref alone while its comment promised a stable order to any future field walking the vector. Two observations can share a ref, so the promise was false.

Round 2's ordering intent is preserved unchanged: containment-first for `customCompliance` and `serviceResults`, time-first for `accessDecisions`. The key only appends.

Cost is bounded: the key sits inside `then_with`, so it is computed only when the named chain ties — never in the corpus (all 16 scenarios carry one row per array). Determinism holds across platforms: `serde_json::to_string` over derived `Serialize` emits declaration order with no map iteration and no floats in these types.

Verified independently: RED reproduces at `1355be13` (1 failed), full crate green at head (2456 passed, 0 failed), no `expected.json` touched.

One item left for your judgment rather than changed unilaterally: `ComplianceCustomEvaluation.named_data` is copied verbatim from a single source fact and never sorted, unlike the merged `named_data`. It is arguably that fact's own content rather than a collector artifact, so it was left alone. If you read it the other way, it is a fourth site.

@coderabbitai full review

adamgell:
Round 4 pushed as `f96a1483`. Both findings were valid and are the same shape as each other: a test that constrains a *relationship* without constraining the *result*.

**Resolved order now pinned.** `facts_agreeing_on_every_named_key_do_not_follow_caller_order` asserted only `forward == reverse`. Reversing the content-key comparison keeps both sides equal while changing what ships. Expected values were read out of the reducer rather than assumed. Mutation-verified: reversing the comparison now fails on "Alpha before Zebra", where before it passed silently.

**`structural_key` content coverage added.** The new unit test pins each `SettingObservation` field individually rather than the struct as a whole, so a field that stops participating fails and names itself. Mutation-verified: truncating the key to 40 chars fails on `grouping_token`.

Worth noting for the record that the resolved-order gap is the same class as the finding you raised two rounds ago about the policy test asserting symmetry without values. That fix was applied to the policy test and not carried to the tests written afterwards, which is exactly how a closed class reopens. Both are now mutation-verified rather than merely present.

Gates: full crate 2457 passed / 0 failed, clippy `-D warnings` clean, `git diff --check` clean (it caught a trailing blank line, fixed before commit), no non-ASCII in added lines, no `expected.json` touched.

@coderabbitai full review

adamgell:
Round 5 pushed as `0a931daa` (RED), `08339705` (fix), `a2e759f7`, `ed0fd9c7`.

**The Major finding was correct, and the sub-question about `raw_text` turned out to be the whole answer.** `timestamp_content_key` is `(raw_text, original_offset, normalized_utc, kind)`, so it compares raw text first and cannot produce chronology: `"07/31/2026 11:50:00"` sorts before `"2026-07-30T00:00:00Z"` because `'0' < '2'`, a full day backwards. Even at a fixed format, an `Offset` stamp reading `2026-07-31T13:50:00+02:00` (11:50Z) sorts after a `Utc` stamp reading `12:00:00Z` despite being the earlier instant, and `IntuneTimestampKind::Offset` exists precisely because sources emit that shape.

So merely dropping the chronology claim would have preserved an order that *looks* chronological on this corpus and silently is not. The reason it stayed invisible is now confirmed: every timestamp in the compliance corpus is exactly `YYYY-MM-DDTHH:MM:SSZ`, and same-shape RFC 3339 sorts chronologically as text by accident.

The comparator now leads with `chronological_key` (the normalized instant), then `timestamp_content_key`, then evidence and the record fields, closing with `content_key`. Unplaceable decisions sort **after** the dated run rather than before it, since `None < Some` would have made them read as "before everything"; within that trailing group `timestamp_content_key` still separates them.

**Compile-time enforcement achieved on the field table**, and I verified it independently rather than accepting the claim: adding a probe field to `SettingObservation` produces `error[E0027]: pattern does not mention field`. A new field can no longer escape the coverage table silently.

**Three further overclaiming comments were found and fixed beyond the ones flagged** - `reduce_access` said a compliance record must *precede* a decision when the filter is `reported <= occurred`; `latest_submission_at` said "a single submission" when the code counts distinct *stamps*; and the new decisions comment initially described only zone-less stamps when the trailing group also holds stampless ones. That class of defect is what every round here has actually been about.

Verified independently: RED reproduces at `0a931daa` with `["a-later", "a-unplaceable", "z-earlier"]` against the expected `["z-earlier", "a-later", "a-unplaceable"]`; full crate green at head; no fixture touched (every corpus scenario holds at most one access decision, so no exported order could change).

@coderabbitai full review

PR #548, opened 2026-08-10
refactor(intune): ask the citation predicate instead of restating it in two lanes

adamgell:
## What was reported

The Autopilot and Compliance lanes were reported as each carrying their own
`is_evidence_backed` predicate.

## What is actually there

The premise is half right. There is exactly **one** `is_evidence_backed` in the
tree (`crates/cmtraceopen-parser/src/intune/evidence.rs:355`). What the two
lanes carried was the hand-inlined De Morgan **negation** of it, inside their
private `finding()` constructors:

```rust
if evidence.is_empty() && coverage_gap_ids.is_empty() { return None; }
```

The other three Windows lanes (Win32, Microsoft Store, Configuration) already
called the shared predicate at their own choke points.

## Arm-by-arm comparison

The guard reads exactly two bits and no lane-specific state at all: it never
consults the snapshot, access state, parse state, severity, or confidence. So
the input space is four cells, and it is exhaustive.

| `evidence` | `coverage_gap_ids` | Autopilot inline | Compliance inline | shared predicate |
|---|---|---|---|---|
| empty | empty | `None` (suppressed) | `None` (suppressed) | `false` |
| empty | non-empty | `Some` | `Some` | `true` |
| non-empty | empty | `Some` | `Some` | `true` |
| non-empty | non-empty | `Some` | `Some` | `true` |

Suppression agrees with the negation of the shared predicate on every cell.

Autopilot normalizes (`normalized_evidence`, plus `sort`/`dedup` on the gap
ids) **before** deciding. That is safe rather than lucky: `sort` never changes
a vector's length and `dedup` never empties a non-empty one, so normalization
preserves both emptiness bits and cannot move the answer.

## Verdict

**The verdict is identical, and identical for the same stated reason.** Each
lane's own doc comment already named `IntuneFinding::is_evidence_backed` as the
invariant it was enforcing, so the agreement is not coincidental, and a lane
emitting an uncited finding would violate the invariant declared on
`IntuneFinding` itself. That is all three parts of the doc's extraction test.

**The constructors hosting it are deliberately per-lane and are NOT unified
here.** Autopilot normalizes both citation sets and takes
`recommended_checks: &[String]`; Compliance keeps citations verbatim and takes
a single `check: &str`; Win32 runs the redaction grammar over `summary` after
the gate; Microsoft Store takes `&[&str]`. A shared constructor would have to
pick one normalization policy and would either impose Autopilot's
byte-identical ordering contract on Compliance or retire it from Autopilot.

The ESP lane carries the same constructor shape at `src/esp/rules.rs:669` over
`EspDiagnosticFinding` / `EspEvidenceRef`, a separate type family that cannot
call this predicate at all. Left alone.

## What changed

Three commits, in this order:

1. **`test(intune)`** pins the current behavior of both lanes across the full
   four-cell space, plus what is *not* shared between the constructors
   (Autopilot sorts and dedupes; Compliance preserves order and duplicates and
   wraps its single check).
2. **`refactor(intune)`** replaces the two inlined negations with
   `candidate.is_evidence_backed().then_some(candidate)`. The pinning tests are
   unchanged by this commit and still pass.
3. **`docs(intune)`** adds the verdict to the "actually shared" table and adds
   protected divergence 9 covering the constructors, so the next consistency PR
   stops at the predicate. All eleven citations are repository-root-relative
   and were checked line by line against what the prose claims.

No behavior change rides along. No `expected.json` fixture is touched. No
genuine defect was found during the work.

## Gates

```
cargo test -p cmtraceopen-parser
  suites: 50  passed: 2457  failed: 0  ignored: 1

cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s

git diff --check
  (no output, exit 0)
```

Stacked on `fix/compliance-sid-redaction` because PRs #545 and #546 already
rewrote `tests/intune_windows_compliance.rs` on that chain.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved finding validation so results are reported only when supported by valid evidence.
  * Preserved citation ordering and duplicate entries consistently.
  * Standardized citation handling across Windows compliance and enrollment checks.
  * Ensured recommended checks retain their associated evidence and coverage.

* **Documentation**
  * Added documentation describing shared evidence-validation behavior and how it varies across Windows assessment areas.

### DIFF
--- a/crates/cmtraceopen-parser/src/intune/apps/windows/microsoft_store/reducer.rs
+++ b/crates/cmtraceopen-parser/src/intune/apps/windows/microsoft_store/reducer.rs
@@ -24,7 +24,7 @@
 //! somebody else's transaction.
 
 use crate::intune::evidence::{
-    IntuneAccessState, IntuneArtifactCoverage, IntuneArtifactStatus, IntuneErrorCode,
+    access_state_for_artifact_status, IntuneAccessState, IntuneArtifactCoverage, IntuneErrorCode,
     IntuneEvidenceRef, IntuneFindingConfidence, IntuneNamedValue, IntuneObservationContext,
     IntuneParseState, IntuneProvenance, IntuneSensitivity, IntuneTimestamp, IntuneTimestampKind,
 };
@@ -91,18 +91,6 @@ pub fn analyze_store_bundle(artifacts: &[StoreSourceArtifact]) -> StoreAnalysis 
 }
 
 // ── Observation extraction ──────────────────────────────────────────────────
-
-fn access_state_for(status: &IntuneArtifactStatus) -> IntuneAccessState {
-    match status {
-        IntuneArtifactStatus::Available => IntuneAccessState::Available,
-        IntuneArtifactStatus::Missing => IntuneAccessState::Missing,
-        IntuneArtifactStatus::PermissionDenied => IntuneAccessState::PermissionDenied,
-        IntuneArtifactStatus::Capped => IntuneAccessState::Capped,
-        IntuneArtifactStatus::Skipped => IntuneAccessState::Skipped,
-        IntuneArtifactStatus::ParseFailed => IntuneAccessState::Failed,
-        IntuneArtifactStatus::Unsupported => IntuneAccessState::Unsupported,
-    }
-}
 
 fn collect_from_artifact(artifact: &StoreSourceArtifact, observations: &mut Vec<StoreObservation>) {
     let Some(payload) = &artifact.payload else {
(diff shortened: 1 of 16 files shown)


adamgell:
Fixed at `c66ed104`. All `.../` placeholders in the divergence-9 section are now repository-relative in `path:line` form.

The sweep also found **two further citations that do not resolve and were not introduced here**: the divergence-7 entry added in #546 cites `intune/device/windows/compliance/redaction.rs` and `tests/intune_windows_compliance.rs` without the crate prefix. Those are on main now, so I corrected them here rather than leaving a known-bad citation on main pending a separate PR.

Worth recording why they survived: my own citation check only validated paths already beginning with `crates/` or `docs/`, so both an abbreviated `.../` path and a crate-relative one were structurally invisible to the check written to catch exactly this. That is the third time this session a verifier has been narrower than the defect it targeted. The check is now prefix-agnostic and additionally rejects any surviving `.../`.

Verified: every path-shaped citation in the document resolves from the repository root, no abbreviations remain, and the cited line numbers point at what the prose claims.

@coderabbitai full review

PR #610, opened 2026-09-10
fix(event-log): load streamed live-channel batches in every path

adamgell:
## Symptom

Opening the Event Log Viewer on Windows showed **no events**, while the backend read the channels in full (Application: 17,833 events on the box, 145 in the last 24 h).

## Root cause

`evtx_query_channels` streams each channel's records as `evtx-record-batch` events and takes those records out of the reply (`std::mem::take` in `event_log/commands.rs`), so **`reply.records` is empty by design** and the reply carries only a count.

Only `queryChannels` drained the streamed batches. The other two live paths merged `reply.records` alone:

- `enumerateChannels` — the auto-load the workspace opens with (the reported case)
- `refreshLoadedChannels` — the time-window refetch

Both showed an empty view while the batches sat undrained in `pendingBatches`. Neither compared the reply's count against what arrived, so a shortfall was silent.

## Fix

- `takeChannelRecords()` takes both sources in one place, and reports a missing batch number or a declared-but-undelivered count as a coverage gap
- `enumerateChannels` and `refreshLoadedChannels` use it; `queryChannels` is refactored onto it with no behaviour change
- both paths reset buffered batches before starting, so a previous attempt cannot be counted towards this one

## Evidence

- **RED first**: 3 new tests in `evtx-store-coverage.test.ts` failed (autoload streaming, refresh streaming, shortfall gap) → 23 passed / 3 failed. GREEN after the fix: 151/151 across `src/workspaces/event-log/`
- `npx tsc --noEmit` clean
- Full suite: 1077/1078 (the one failure is the pre-existing flaky `EspDiagnosticsWorkspace` workload-table render, which times out at the 5 s default and passes on rerun — untouched by this diff)
- **Native**: `cargo run --example evtx_scan --features event-log -- --days 1 --channel Application` returned 145 events on the same box, proving the query path was fine
- **In-app**: after the change, the running dev app loaded the channels and events are visible

Part of #539. Blocks EVTX-002 (live channel browse) validation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event log loading and refreshing to reliably combine streamed records with query results.
  * Added validation to detect missing event batches and incomplete event counts.
  * Prevented stale streamed data from being carried into new loads or refreshes.
* **Tests**
  * Added coverage for streamed event batches during automatic loading and refresh operations.


adamgell:
Superseded by the canonical streamed-record implementation shipped in PR #583 (merged as superseded, 481e5fe5), squash-merged to main as c77c1f10. Branch history preserved locally in a verified git bundle before remote branch deletion.

PR #609, opened 2026-09-10
feat(event-log): mark Event Log Viewer as preview

adamgell:
## Summary

Two-surface preview badge for the Event Log Viewer workspace (Adam-approved design, 2026-09-10 handoff):

- **Registry label suffix** — `eventLogWorkspace.label` becomes `"Event Log Viewer (Preview)"`, rendering automatically in the workspace dropdown and status bar
- **Filter-bar ghost badge** — Fluent `Badge` (`appearance="ghost"`, `size="small"`) labeled `Preview` leads the filter bar's first row, persistent through file mode, live mode, filtering, grouping, and detail-pane states
- The empty-state `SourcePicker` title is deliberately NOT badged (a badge that vanishes on first data load was rejected in design review)

## Testing

- Focused Vitest + Testing Library following `EventLogWorkspace.test.tsx` conventions: registry label contract and loaded-state badge presence
- `npx vitest run src/workspaces/event-log/EventLogWorkspace.test.tsx` — 6/6 passed
- `npx tsc --noEmit` — clean
- `npm test` — 105 files, 1077 tests passed

Part of #539 (preview badge only; epic completion requires further Windows evidence)

## Summary by CodeRabbit

* **UI Updates**
  * Added a non-interactive “Preview” badge to the event log filter toolbar.
  * Updated the event log workspace label to “Event Log Viewer (Preview)”.
  * Updated the status bar and active-view badge to display the event log’s preview label consistently.
  * Preview indicators now appear when event log content is loaded.

PR #611, opened 2026-09-10
fix(event-log): bound the automatic load and fill the view as channels land

adamgell:
## Symptom

Opening the Event Log Viewer left the workspace empty and its controls disabled for minutes. On a measured Windows machine:

| Channel, last 24 h | Events | Read time |
|---|---|---|
| Security | 239,704 | **318 s** |
| Application | 152 | 1.9 s |

The window alone never bounded a channel, and the workspace waited for the slowest one.

## Root cause

1. The automatic load and the time-window refetch asked for every event in the window (`maxEvents: null`), so Security alone was a five-minute read.
2. `queryChannels` collected **all** channel promises before merging any of them — a bulk selection showed nothing until the slowest channel answered.
3. One request per channel was issued for every selected channel at once, so a selection of every channel on the machine queued hundreds of reads at a service that serializes much of that work.
4. Nothing stopped a remount from starting a second scan of every channel; three concurrent Security scans were observed contending on 8 vCPUs.

## Fix

- **Bounded automatic load**: the automatic load and the time-window refetch ask for the newest **2,000** events per channel. The explicit **Load** stays unbounded, so a whole channel remains reachable.
- **Honest disclosure**: a channel that comes back at the bound is reported in the coverage banner — `newest 2000 events shown — the channel may hold more`. This side cannot tell whether the channel held more (the read stopped at the bound), so the note does not claim it does.
- **Bounded fan-out**: a bulk selection reads **4** channels at a time instead of all of them at once.
- **Fill as it lands**: each channel is merged the moment it arrives, so the first channel is visible immediately instead of after the slowest one.
- **One automatic load**: a second mount joins the load already running rather than scanning every channel again. The guard is cleared in `finally`, so a failed load cannot wedge every later mount.

## Evidence

- **RED first**, five behavioural failures: `[null] != [2000]` (uncapped), no cap disclosure, `[null, null]` length 2 (two stacked automatic loads), `12 != 4` (unbounded fan-out), and a fast channel not visible while a slow one is gated.
- **GREEN**: 32/32 in `evtx-store-coverage.test.ts`; 11 files / 157 tests across `src/workspaces/event-log/`; `npx tsc --noEmit` clean.
- Measurements come from the native scan path on the same machine (`cargo run --example evtx_scan --features event-log -- --days 1 --channel Security`).

## Stacking

Based on **#610** (`fix/event-log-streamed-batch-intake`), which this builds on: on `main` those two load paths merge only the reply, which the streaming backend empties, so bounding them without #610 would bound a path that shows nothing.

## Not in this change

`Load` and `Refresh` stay disabled while a load runs. Ungating them needs per-operation state — Refresh clears records while another load can be mutating the same store — so it is left for a design with its own tests rather than enabled here.

Part of #539.


adamgell:
## Consolidation checkpoint — this PR cannot be merged as-is

The Event Viewer candidate has been consolidated onto current `main` plus the `92752b21` repair/export line. **This PR is built on `#610`'s `takeChannelRecords` helper and the pre-`requestId` `resetStreamedRecords`, neither of which exists on that candidate**, so it is not integrable by merge resolution.

### Evidence that the semantics are still needed

Nothing here is superseded; the canonical tree has none of it:

- no `AUTO_LOAD_MAX_EVENTS` (or any per-channel cap) — the automatic load is unbounded;
- no concurrency limit on channel queries;
- no `autoloadInFlight` coalescing of equivalent requests.

### Evidence that the patch cannot be transplanted

Its conflicts are in `evtx-store.ts` and its coverage test, and the mismatch is not cosmetic: it calls `resetStreamedRecords(availableCore)` while the candidate's signature is `resetStreamedRecords(channels, requestId)`. The helper it introduces sits on top of `#610`, which the candidate superseded — the streaming line already resets, awaits reconciliation, drains and acknowledges on every loading path. Transplanting this patch would mean re-implementing it against the current request-ID, streaming, reconciliation, acknowledgement, cancellation and stale-result infrastructure, which must not be guessed at from a diff.

### Canonical base for the replacement

`orchestration/event-viewer-epic` @ **`7c166660`** — current `main` (v1.6.0), the `92752b21` repair/export line, #609 Preview labelling, and an isolated Clippy cleanup. The workspace passes `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` at that commit.

### Sequencing

This replacement comes **after** the #613 replacement, which supplies the channel-eligibility semantics that this one consumes: disabled channels must not enter the automatic fan-out, while unknown channels stay eligible because probing fails open. Branch from the candidate as it stands once that replacement is in — not from here.

### What replaces it

A fresh branch owned by this PR's author, RED-first, proving the behaviour below against the current infrastructure. This PR stays **open** as the behaviour contract and the historical record — measurements, discussion and tests included — and will be closed as superseded once the replacement has committed RED tests and an implementation plan. This PR's commits will **not** be forced into the new ancestry.

Handoff, as decided:

> Reimplement #611 on top of the canonical branch containing the replacement #613 semantics. Do not restore `takeChannelRecords`; use the current request-ID, streaming, reconciliation, acknowledgement, cancellation and stale-result infrastructure. Start RED-first and prove the 2,000-per-channel automatic limit, four-channel concurrency, progressive display, equivalent-request coalescing, changed-request separation, stale-result rejection, honest cap coverage and explicit unbounded Load. Disabled channels must not enter automatic fan-out. Return the RED evidence and implementation plan before changing production behaviour.

adamgell:
## Replacement landed on the consolidated candidate

The behaviour this issue asks for has been rebuilt on the current canonical head, RED first. The
original patch could not be transplanted: it calls `resetStreamedRecords(channels, requestId)`
against an ingestion API that no longer exists, and it depends on `#610`'s `takeChannelRecords`
helper, which is not on that candidate. Nothing was hand-ported from the old diff; the behaviour was
re-derived against the store as it stands.

**Where it lives:** `fix/event-log-autoload-bounds`, integrated into
`orchestration/event-viewer-epic` (epic PR #583). Its base is the integrated `#613-v2` head, so the
two replacements are stacked in the order requested.

- `edd7e063` integrated canonical head (`#613-v2`, channel eligibility tri-state)
- `0d7f0b4b` RED: 12 tests, 5 failing for the stated reasons
- `6cd82155` lane met its base
- `be6150e3` implementation: 13 of 13 green

### What it does now

- `AUTO_LOAD_MAX_EVENTS = 2000` bounds the newest slice on the automatic enumeration and on the
  time-window refetch. The explicit Load stays unbounded, so a whole channel remains reachable.
- The automatic fan-out runs through the existing four-read pool that already bounded the bulk
  paths, instead of one unbounded read per channel.
- `enumerateChannels` coalesces by scope, so a remount while a scan is reading no longer starts a
  second full scan. Joining also requires the running request to still be current. That second
  condition is load-bearing: scope-only joining lets a mount wait on a scan a newer load already
  superseded, which publishes nothing. Reverting it to scope-only fails 5 of the 13 tests.
- A capped channel says so: `"<channel>: newest 2000 events shown (the channel may hold more)"`.
  It is phrased as a slice, never as a channel that was read in full, because a channel that came
  back at the bound may hold more and the read side cannot tell.

### Constants

Both are choices, not measurements, and the code says so. `2000` is picked for a busy channel's
recent activity; the Windows figures quoted in the original PR (Security 239,704 events in 24 h,
318 s) are not reproducible on this machine, so they justify bounding the read but do not pin the
number. `4` is the value this tree already used.

### Two pre-existing tests were adapted, neither weakened

`does not let a superseded enumeration finish a newer load` triggered its supersede with two
identical-criteria calls, which coalescing now makes one load; the trigger changed to a different
filter and its stale-write assertions are untouched. The thousand-channel refetch test pinned
`maxEvents === null` on every call, which is the unbounded contract this replaces, so that pin is
deleted rather than re-pointed at the new number; the bound is asserted by name on both paths in the
new file.

### Verification on `be6150e3`

New file 13 passed (5 red before the change). Event-log workspace 30 files / 473 passed, previously
460. Full frontend 131 files / 1,543 passed. Full Rust (`src-tauri`) 1,554 passed / 0 failed / 6
ignored; Lite 402 / 0; parser crate 2,752 / 0 / 1 ignored. Workspace and `x86_64-pc-windows-msvc`
Clippy clean under `-D warnings`. `tsc` exit 0. The join condition was mutation-tested: reverting it
to scope-only fails exactly the 5 tests claimed, and the tree was restored byte-identical.

The Windows-lab items this issue originally listed are unchanged and still unobserved: that the
service reports switched-off channels in a live census, that `EvtQuery` on a switched-off channel
fails, and remote-session probing against a second machine.

Issue left open: the fix is on the epic branch, not on `main`.

adamgell:
Superseded by be6150e3 (fix/event-log-autoload-bounds), shipped in PR #583 and squash-merged to main as c77c1f10. Branch history preserved locally in a verified git bundle before remote branch deletion.

PR #613, opened 2026-09-11
fix(event-log): stop treating disabled channels as coverage gaps

adamgell:
## Symptom

Loading every channel filled the coverage banner with hundreds of gap lines, e.g.

```
AirSpaceChannel: EvtQuery(AirSpaceChannel): The request is not supported.
AMSI/Debug: EvtQuery(AMSI/Debug): The request is not supported.
```

## Cause

`Select all` selected every enumerated channel, including the ones that are switched off, and the service refuses a disabled channel. Each refusal was reported as a coverage gap, so the banner described the *machine's configuration* as though the view were missing events that exist.

Measured on the machine the gaps came from:

| | |
|---|---|
| Channels enumerated | 517 |
| Enabled | 432 |
| Disabled, **no log file** | 85 |
| Enabled channels with events | 169 |
| …of those, readable | **169 (0 failures)** |

`AirSpaceChannel` is type `Analytical`, `enabled: false`, and its log file does not exist. There was nothing to read; the report was noise.

## Fix

- **Enumeration reports whether each channel is recording** (`EvtOpenChannelConfig` + `EvtGetChannelConfigProperty(EvtChannelConfigEnabled)`), so the UI knows which channels can hold events.
- **Select all selects only those channels.** Disabled channels stay in the tree and are marked `disabled` — visible, not hidden, which is why the selection leaves them out.
- **The probe fails open**: an unreadable configuration counts as enabled, so a failed probe can never hide a live channel.
- Genuine failures (a channel that *is* recording but cannot be read, e.g. denied without elevation) still become coverage gaps. On this machine that set is empty.

The epic's rule is preserved, not relaxed: a channel that holds events and cannot be read is still reported; a channel that holds nothing is not described as a hole in the view.

## Evidence

- **RED first**: `selectAllChannels` returned `['Application','AirSpaceChannel']`, expected `['Application']` → GREEN after the fix.
- Frontend: `npx tsc --noEmit` clean; 11 files / 149 tests across `src/workspaces/event-log/`.
- **Native, this machine**: `a_channel_whose_configuration_cannot_be_read_counts_as_enabled` ok, `disabled_channels_are_identified_among_the_enumerated_ones` ok, `live_query_application` ok — 3 passed, 0 failed.
- Census above from `Get-WinEvent -ListLog *` plus a per-channel read attempt.

Part of #539.

## Summary by CodeRabbit

- **New Features**
  - Event Log channels now indicate whether they are enabled and readable.
  - Disabled channels remain visible and are labeled accordingly.
  - File-based channels continue to appear as enabled.
- **Bug Fixes**
  - Bulk channel selection now excludes disabled channels while retaining enabled channels.
  - Channel status is reported consistently when channel scanning or queries fail.


adamgell:
## Consolidation checkpoint — this PR cannot be merged as-is

The Event Viewer candidate has been consolidated onto current `main` plus the `92752b21` repair/export line. **This PR's implementation targets an ingestion API that no longer exists on that candidate**, so it is not integrable by merge resolution.

### What was tried, and what it showed

Merging `042d47bb` into the candidate conflicted in five files — `src-tauri/src/event_log/{commands,live,parser}.rs`, `src/workspaces/event-log/types.ts` and `evtx-store-coverage.test.ts` — and none of those conflicts were the kind that a resolution can settle honestly.

### Evidence that the semantics are still needed

Nothing here is superseded; the behaviour is absent from the canonical tree:

- no `EvtChannelConfigEnabled` call anywhere;
- no `channel_is_enabled` function;
- no `enabled` field in `src/workspaces/event-log/types.ts` or the IPC DTO.

### Evidence that the patch cannot be transplanted

The change was written against `c90481d9`, before the streaming line reworked channel loading. Its ingestion helpers and call shapes no longer exist in that form, so adapting it means re-implementing against the current API rather than resolving conflicts — which is exactly what must not be guessed at from a diff.

### Canonical base for the replacement

`orchestration/event-viewer-epic` @ **`7c166660`** — current `main` (v1.6.0), the `92752b21` repair/export line, #609 Preview labelling, and an isolated Clippy cleanup. The workspace passes `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` at that commit.

### What replaces it

A fresh branch from `7c166660`, owned by this PR's author, with #613's behaviour reimplemented RED-first against the current reconcile/drain APIs. This PR stays **open** as the behaviour contract and the historical record — measurements, discussion and tests included — and will be closed as superseded once the replacement has committed RED tests and an implementation plan. This PR's commits will **not** be forced into the new ancestry; only their semantics carry forward.

Handoff, as decided:

> Reimplement #613 from a fresh branch based on the canonical Event Viewer head after the isolated Clippy cleanup. Treat this PR as the evidence and behaviour contract, not as a patch to transplant. Start with RED tests for enabled, disabled, unknown/probe-failed, and enabled-but-unreadable channels. Preserve fail-open configuration probing, keep disabled channels visible but outside Select All/default acquisition, and retain genuine acquisition failures as coverage gaps. Do not implement automatic-load limits or scheduling; those belong to #611. Report the branch/SHA, RED results, changed files, native API contract and smallest implementation plan before broad production changes.

adamgell:
Superseded by #634, which carries this behaviour reimplemented against the current ingestion API.

The replacement is established as the process required: RED tests committed at `ab296ab9` on `fix/event-log-channel-eligibility` (9 failing, 87 pre-existing passing, `tsc` clean) plus the implementation plan, the native API contract read out of the pinned `windows` crate, and the list of what only Windows can establish.

This PR remains the behaviour contract and the historical record — the measurement (432 enabled / 85 disabled channels on a real machine), the discussion, and the four review-driven tests CodeRabbit asked for here are all carried into the replacement. Its commits are deliberately **not** in the new ancestry: they target an API this branch no longer has, and forcing them in would break the build.

Worth recording for whoever reads this later: the native assertion on this PR was `disabled < channels.len()`, which passes when no channel is reported disabled at all — the check could not see the defect class it existed to prevent. The replacement asserts `disabled > 0` against `Get-WinEvent -ListLog *`, and requires the `EVT_VARIANT` type to be checked before reading the boolean, since this patch read the union unchecked and a zeroed variant would have read as "switched off" — the opposite of fail-open.

adamgell:
Superseded by edd7e063 (fix/event-log-channel-eligency, PR #634) and covered on main via PR #583 (c77c1f10). Recording links for posterity; issue was already closed. Branch history preserved locally in a verified git bundle before remote branch deletion.

Issue #549, opened 2026-08-10
ESP session export writes an unredacted snapshot to a user-chosen file (redaction projection exists but is never called)


adamgell:
## Summary

The ESP Diagnostics workspace's Export button writes the **raw** snapshot to a user-chosen JSON file. No redaction runs on that path. A purpose-built projection for exactly this (`crates/cmtraceopen-parser/src/esp/redaction.rs`, `redacted_export_projection`) exists, is tested, and is **never called by the application**.

## Verified call path

```
src/workspaces/esp-diagnostics/EspDiagnosticsWorkspace.tsx:328  exportSession()
  :329   const current = useEspDiagnosticsStore.getState().snapshot     <- raw store snapshot
  :339   buildEspSessionCapture(current, ...)
           -> src/workspaces/esp-diagnostics/esp-session-capture.ts:38-44   { ..., snapshot }   (embedded unmodified)
  :342   writeTextOutputFile(destination, serializeEspSessionCapture(capture))
           -> esp-session-capture.ts:48   JSON.stringify(capture, null, 2)
           -> src/lib/commands.ts         invoke("write_text_output_file", ...)
           -> src-tauri/src/commands/file_ops.rs:481-487   fs::write(&path, contents)   (verbatim)
```

No redaction at any step.

## Scope of the problem is wider than one button

An audit of every call site shows **zero** invocations of any lane's `redacted_export_projection` from `src-tauri/` or `src/`. All 20 references live inside `crates/cmtraceopen-parser/src/` and are that module's own tests plus re-exports. The only place a projection is applied by construction anywhere in the codebase is `crates/cmtraceopen-parser/src/intune/portal/windows/company_portal/logs/document.rs:30-34`, where `parse_log_document` *is* the projection wrapping the preserving variant.

So the redaction work across the Intune and ESP lanes is, as shipped, library surface that nothing reaches.

## Why this is worse than a plain missing call

The UI already masks these values on screen. `src/workspaces/esp-diagnostics/esp-view-model.ts:14-15` states the disclosure policy and `displayEvidenceValue` (:126-137) enforces it for **rendering only**. `exportSession` reads the store directly and never touches the view model.

A user therefore sees masked values on screen, clicks Export, and gets cleartext. The on-screen masking actively creates the false impression that the export is protected.

## Related, same root cause

[...]

PR #617, opened 2026-09-14
fix(intune): scrub masked identity literals out of Autopilot free text (#564)

### LINKED ISSUE #564: Autopilot export leaves bare serials, tenant domains, and device names in free text
## Summary

The Autopilot export projection masks device and tenant identity where those
values are **typed fields**, but leaves the same literal values intact when they
appear inside **free narrative text** (observation messages, finding summaries,
timeline detail, named-data values).

This is the same class of gap that PR #558 closed for the ESP export under #549.
It was named there and deliberately left out of scope, because it is a
redaction-policy change rather than a boundary change.

## Evidence

`autopilot::redact_text` is the free-text rule set. Its rules are a UPN regex, a
`\Users\<name>` path segment, and an opaque Base64 blob of 40+ characters. A
bare serial, a bare DNS tenant domain, and a bare device name match none of
them:

```rust
// crates/cmtraceopen-parser, probe against the public API
let narrative = "device 5CD9SYNTH01 joined contoso.onmicrosoft.com as DESKTOP-SYNTH1";
autopilot::redact_text(narrative);
// -> "device 5CD9SYNTH01 joined contoso.onmicrosoft.com as DESKTOP-SYNTH1"
```

Meanwhile the same values *are* masked where they are typed, in
`redacted_export_projection`:

```rust
projected.identity = AutopilotDeviceIdentity {
    serial_number: redact_opt(&snapshot.identity.serial_number),
    tenant_domain: redact_opt(&snapshot.identity.tenant_domain),
    device_name:   redact_opt(&snapshot.identity.device_name),
    ...
```

So an export can carry `identity.serialNumber = "[redacted:a1b2…]"` and, two
fields away, an observation message naming the serial in the clear.

A bare serial has no distinctive shape and a bare DNS domain has no label, so no
pattern rule can close this. `apps/windows/common::redact_text` has the same
property for the same reason: it masks `ComputerName: DESKTOP-AB12CD` because
the label is there, and leaves the identical value unlabelled in narrative.

## Suggested approach

The mechanism ESP now uses (`crates/cmtraceopen-parser/src/esp/redaction.rs`):
collect the literal values of the typed fields the projection is *about to
mask*, then scrub those exact strings out of every free-text field. The
projection already holds the values, so the gap closes by construction rather
than by another pattern.

Two details that mattered there and will matter here:

- **Order.** The literal scrub must run *last* in each free-text pipeline. The
  tenant domain is a substring of the UPN, so scrubbing it first breaks the
  mail-address match and leaks the local part.
- **A length floor.** Firmware reports junk serials (`0`, `N/A`, `None`). A
  value that short cannot be told apart from an ordinary word once it sits
  unlabelled in narrative. ESP skips literals under six bytes, staying below a
  seven-character Dell service tag.

Autopilot can do this **better than ESP did**. Its masks are `stable_token`
values derived from the value itself, so a free-text occurrence can be replaced
with the *same* token as the typed field. Correlation survives the scrub instead
of collapsing to a flat `[redacted]`, which is exactly the property the module
docs already argue for.

## Scope

Checked every module with a `redacted_export_projection`:

- **Affected:** `intune/enrollment/windows/autopilot` — the only sibling family
  whose snapshot carries typed device/tenant identity (`serial_number`,
  `tenant_domain`, `device_name`, `hardware_hash`, `user_principal_name`).
- **Not affected:** `apps/windows/{win32,microsoft_store,scripts,remediations}`,
  `device/windows/{compliance,configuration}`, and the Company Portal families
  (Windows, macOS, iOS/iPadOS). Their models carry no typed device-identity
  field, so there is no literal to collect. Their shared free-text helper has the
  same unlabelled-value blind spot, but nothing in those snapshots supplies a
  value to scrub with.
- **Already closed:** `esp` (PR #558 / #549).

Adjacent, not part of this issue:

- `dsregcmd/models.rs` has a `tenant_domain` and no redaction or export
  projection at all. Worth confirming this gap is considered when the DsRegCmd
  export lands (#556).
- `intune/models.rs` has a `device_name` with no masking on any path.

## Acceptance

- A bare serial, tenant domain, and device name planted unlabelled in an
  Autopilot observation message do not survive `redacted_export_projection`.
- Free-text occurrences mask to the same `stable_token` as the typed field, so
  the export still shows that two records name the same device.
- A degenerate short serial does not scrub unrelated narrative.
- Idempotence holds: the projection of a projection serializes identically.

adamgell:
Closes #564 once merged.

## The gap

`redacted_export_projection` masked `serial_number`, `tenant_domain`, `device_name`, `hardware_hash` and `user_principal_name` where they are typed fields, and masked nothing when the very same literal appeared in narrative text. `autopilot::redact_text` knows a UPN regex, a `\Users\<name>` path segment and a 40+ char Base64 blob, so a bare serial or a bare DNS domain matched no rule at all.

Concretely, before this change an export could carry `identity.serialNumber = "[redacted:a1b2…]"` and, two fields away, an observation message naming that serial in the clear.

## The mechanism

The projection collects the literal values it is about to mask, pairs each with the `stable_token` its typed field carries, and scrubs those exact strings out of every free-text field. Two properties are load-bearing and both are pinned by tests:

- **The scrub runs last**, behind the shaped rules. The tenant domain is a substring of the UPN, so scrubbing it first would consume the mail-address match and leave the local part in the clear. Inverting the pipeline makes `a_tenant_domain_literal_does_not_break_the_upn_match_that_precedes_it` fail with `synthetic.user@[redacted:…]`.
- **Literals under six bytes are skipped.** Firmware reports junk serials (`0`, `N/A`, `None`) that cannot be told from an ordinary word in prose. Removing the floor makes `a_degenerate_short_serial_does_not_scrub_unrelated_narrative` fail by tokenizing `N/A` in plain narrative.

This is the mechanism ESP adopted for #549, with one improvement the Autopilot model allows: because its masks are `stable_token` values derived from the value itself, a free-text mention is replaced with the *same* token as the typed field, so the export still shows that two records name the same device instead of collapsing both to a flat marker.

## Evidence

```
cargo test -p cmtraceopen-parser --test intune_windows_autopilot
  test result: ok. 52 passed; 0 failed; 1 ignored      (the ignored test is the pre-existing golden rewriter)

cargo test -p cmtraceopen-parser --lib -- autopilot::redaction
  test result: ok. 9 passed; 0 failed

cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings
  clean
```

RED before the change:

```
thread '...a_free_text_literal_masks_to_the_same_token_as_its_typed_field' panicked:
  the free-text mention must carry the typed field's token [redacted:44dba3900fa96e94],
  got ... Device SYNTH-DEV-01 (SYNTH-SERIAL-0001) last checked in to contoso.example.
thread '...bare_identity_literals_in_narrative_do_not_survive_the_export' panicked:
  the bare literal "SYNTH-SERIAL-0001" survived the exported projection
```

## Deliberately not done

- No new regex for unlabelled values — no pattern rule can close this, which is why the literal scrub is used instead.
- Sibling families (`apps/windows/{win32,microsoft_store,scripts,remediations}`, `device/windows/{compliance,configuration}`, the Company Portal families) are untouched: their models carry no typed device-identity field, so there is no literal to collect.
- `AutopilotDocumentReport.detail` is left alone: the repository documents it as a reducer-authored sentence whose only source-derived interpolation is the declared document kind, and that kind is exported unredacted next to it.
- A UPN mentioned in prose keeps the shaped rule's `[upn:…]` token rather than the typed field's token; the correlation criterion in the issue is stated for the bare serial/domain/device-name case, which does use the typed token.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection of sensitive identifiers in unstructured Autopilot text.
  * Ensured serial numbers, tenant domains, device names, account identifiers, and hardware hashes use consistent masking across related fields.
  * Added case-insensitive masking for non-ASCII identifiers.
  * Preserved short values that do not meet the masking threshold.
  * Improved handling of conflicting and overlapping sensitive data.
  * Ensured repeated redaction produces stable, consistent results.
  * Prevented account information from being exposed when masking tenant domains.


adamgell:
@coderabbitai review

Finding accepted and fixed, with the ordering property kept intact.

Rather than reordering the pipeline, the shaped rules are now value-aware: the literal table is a `BTreeMap<literal, token>`, and the UPN, user-path and opaque-blob replacements resolve through `shaped_token`, which returns the literal's own typed token when the match is one the export masks and the rule's own token otherwise. A collected UPN or hardware hash therefore now carries `[redacted:…]` in narrative as well as in the typed field, so one identity is one token.

Shaped-rules-first, literal-scrub-last is unchanged, so the tenant-domain hazard stays closed: a tenant domain scrubbed before the shaped rules would consume the UPN match and leak the local part.

Red before the fix: `a_narrative_upn_masks_to_the_token_its_typed_field_carries` failed with `[upn:8d11eba4d0094856]` where `[redacted:8d11eba4d0094856]` was required, and the hardware-hash test failed the same way on `[blob:…]`.

Green: `cargo test -p cmtraceopen-parser` (51 targets, 2649 passed, the two new tests included), `cargo test -p cmtraceopen-parser --lib -- autopilot::redaction` (9 passed), `cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings` clean. The ordering test `a_tenant_domain_literal_does_not_break_the_upn_match_that_precedes_it` and the six-byte-floor test both still pass.

adamgell:
Consistency follow-up on the Autopilot lookup, pushed after your approval at b5c3c077.

Your follow-up pass on the sibling DsRegCmd lane found that its literal scrub folded case ASCII-only, so a classified literal containing a non-ASCII letter did not match a differently-cased occurrence in free text. `MaskedLiterals` had the same shape: an exact trim plus `to_ascii_lowercase` key. A device name carrying a non-ASCII letter, written in narrative with different casing, would not have been scrubbed, and no shaped rule covers a bare unlabelled name.

Both sides of the lookup are now Unicode-aware. The scrub no longer folds a copy of the text, because `to_lowercase` can change a string's length and a folded copy's offsets cannot index the original; it scans the original value with per-character case comparison and returns real byte offsets, so the existing slicing stays valid. Documented limitation: per-character folding reaches one-to-one mappings (`É`/`é`, `Ä`/`ä`) but not the few mappings that change a character's count (`İ` → `i` + combining dot), which are different text rather than another casing. The stored token is still exactly the typed field's token, so correlation, the six-byte floor, the shaped-rules-first ordering and idempotence are all unchanged.

Red first: `a_non_ascii_literal_masks_its_differently_cased_occurrence_to_the_same_token` failed with the variant still visible.

Green: `cargo test -p cmtraceopen-parser --test intune_windows_autopilot` 55 passed (1 ignored), `cargo test -p cmtraceopen-parser` 2651 passed with no failures, `cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings` clean.

Since this lands after your approval, approval at the previous head no longer covers it — a fresh cycle is expected, and CodeRabbit is rate-limited for a while, so no rush.

adamgell:
@coderabbitai review

Fixed exactly as proposed, and the collision edge you implied is gone rather than papered over.

`mask_value` now hashes `value.trim().to_lowercase()`. `MaskedLiterals::new` derives its token from that same canonical string and uses it as the map key, so the table and the typed fields share one canonical form by construction instead of agreeing by coincidence — the previous "first value in byte order wins on collision" behaviour is removed.

Red first: `one_identity_mints_one_token_whatever_case_a_non_ascii_letter_arrives_in` failed with `[redacted:316fe28efa2a71d8]` against `[redacted:19d8b5a2d465d7b8]` when the old `to_ascii_lowercase` token behaviour was restored. The test pins that `PC-ÉLODIE` and `pc-élodie` mint one token, that both narrative spellings carry it, and that neither spelling survives.

Green: `cargo test -p cmtraceopen-parser --test intune_windows_autopilot` 56 passed (1 ignored), `cargo test -p cmtraceopen-parser` 2652 passed with no failures and no existing test updated, `cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings` clean.

One residual I have deliberately left alone and want on the record: the reducer's `distinct_values` still groups case-insensitively with `to_ascii_lowercase`, so two values differing only in the case of a non-ASCII letter are two distinct values there while the export folds them to one token. The visible effect is a conflict whose detail says "2 distinct values" while both exported values are identical tokens — over-reporting, not hiding. Folding it there would change which conflicts the reducer reports, which is a diagnostic-semantics decision rather than a redaction one, so it is not in this PR.

adamgell:
@coderabbitai review

Fixed, and the limitation your finding names is now closed rather than documented away.

The matcher no longer compares one source character at a time. The scrub builds one folded view of the free-text value per call and compares sequences of folded characters, one folded character per side per step, so `İSTANBUL-PC` folds to `i` + U+0307 + `stanbul-pc` and the narrative spelling folds to the same sequence. Every folded character carries the byte range of the source character it came from, and a match reports the range its first and last folded characters cover, so masks still slice the original string and never a folded copy that lowercasing may have resized.

Caseless equality also accepts equal `to_uppercase` expansions, which is what reaches final sigma: `Σ` folds to `σ`, `ς` keeps its own shape, and both uppercase to `Σ`. Lowercase stays the only fold for the key, the token and the match, so one identity still has one canonical form.

Red first: `a_literal_that_folds_to_several_characters_matches_its_other_casing` failed with the narrative still carrying `İstanbul-PC` in the clear.

Green: `cargo test -p cmtraceopen-parser --test intune_windows_autopilot` 57 passed (1 ignored), `cargo test -p cmtraceopen-parser` 2653 passed with no failures and no existing test updated, `cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings` clean.

Two limits remain, and they are inherent to a per-step alignment rather than shortcuts, so I am stating them instead of claiming completeness: the Turkic dotless `i` still does not equal `İ` (default folding keeps the combining dot, and the Turkic mapping would be wrong for this audience), and a one-to-many spelling such as `ß` against `SS` or `ﬁ` against `fi` cannot align when each step consumes exactly one folded character per side. Both are written into the matcher's doc comment.

adamgell:
Documented and pinned rather than normalized, as the finding asked.

The limitations block on `find_ignore_case` now names three classes: the Turkic dotless `i`, the one-to-many spellings (`ß` vs `SS`, ligatures), and the normalization-form gap you named — a typed precomposed `PC-ÉLODIE` folds to `pc-élodie` while a narrative `E` + U+0301 folds to `e`, U+0301 and stops aligning at that position.

It also records why normalization is not performed: it rewrites the narrative on its way into the export, which is a behaviour change with its own trade-offs rather than a free win. A new test pins the gap — the precomposed spelling in another casing is still masked to the typed token, which is what stops the test passing by the matcher merely failing to fold, while the decomposed spelling is asserted readable byte-for-byte. Its comment states that anyone who adds normalization must delete that test deliberately, so the gap cannot be closed silently later.

`cargo test -p cmtraceopen-parser` 2654 passed, no failures; `cargo clippy -p cmtraceopen-parser --all-targets -- -D warnings` clean.

adamgell:
@coderabbitai review

No review has run against the current head (`ee046f6d`) since the last push, and the recorded decision is from an earlier head. All findings from the previous cycle are addressed or dispositioned with zero unresolved threads; a fresh pass at this head is what the gate needs.

adamgell:
## Superseded by #632 — this branch adds nothing

Verified against `origin/main` at `34e24ff0`.

`899bfb31` — `refactor(parser): keep one caseless matcher, in the shared grammar (#632)` — is the squash of `refactor/caseless-grammar`, and its body carries **this PR's own commit**:

```
* fix(intune): scrub masked identity literals out of Autopilot free text (#564)
```

That is this PR's feature, and it is on `main`.

The two files this branch changes are older revisions of the files `main` now carries: `main`'s `autopilot/redaction.rs` is written against the shared `caseless_key` matcher that #632 consolidated, and `main`'s `intune_windows_autopilot.rs` test file already carries this branch's cases plus 51 lines it lacks. Every file this branch touches is therefore either already identical to `main`'s version or an older revision of it, so resolving both to `main`'s version leaves no content this branch alone adds.

Why this still reports `CONFLICTING`: the conflict is semantic, not mechanical. `main` landed #632, which moved the caseless matcher into the shared grammar (`intune/apps/windows/common/redaction.rs`, re-exported from `common/mod.rs`), and this branch predates that split.

**There is no implementation work left here.** Recommendation: close this PR as superseded by #632 rather than merging `main` forward, which would only produce a no-op commit.

Reproduce:

```
git log origin/main --oneline --grep="scrub masked identity"
# 899bfb31 refactor(parser): keep one caseless matcher, in the shared grammar (#632)

git show -s --format=%B 899bfb31 | Select-String '#564'
```

Filed during the convergence pass; see also the companion finding on #620, which is superseded by the same commit.

adamgell:
Closed as superseded by #632 / 899bfb31.

The feature commits from this PR are present in the #632 squash, and current
main contains the resulting implementation. Resolving this branch against main
would add no new product behavior.

PR #618, opened 2026-09-14
feat(app): open a second launch's files in the window that is running (#565)

### LINKED ISSUE #565: Open log files in current cmtraceopen-window
Great tool!

However a major drawback for me is the inability to open additional logfiles in the already open window by double clicking the file or passing commans line arrguments to cmtraceopen. 

Regards
Tobias

adamgell:
Closes #565 once merged.

## Problem

A second launch — double-clicking a `.log` through the file association, or passing a path on the command line — started a second process. The user ended up with a new window holding one tab while the window they were working in drifted away behind it.

## Change

- `tauri-plugin-single-instance` is registered **first** in `src-tauri/src/single_instance.rs`, so a second launch is detected while the app is still being built and exits before `tauri.conf.json`'s empty window can appear.
- The second launch forwards its argv. The running instance parses it with the same `parse_initial_launch_arguments` the first launch uses (now `pub(crate)`), so a positional file path has one definition and an app-owned option can never arrive as a file to open.
- The running instance raises and focuses `main` and emits exactly one `second-launch-open` event; the frontend opens each forwarded path through the flow it already uses for a dropped file, so a forwarded file lands in a tab and in Recent like any other open.
- An elevated relaunch is exempt: it starts while the process it replaces is still holding the single-instance lock, so routing it into the closing window would leave the user with nothing on screen. `openPathForActiveWorkspace` now takes its trigger as a parameter so a forwarded launch is not logged as a drop.

The plugin was chosen over hand-rolled IPC because the repository had no single-instance handling anywhere and the plugin is the established implementation of exactly this.

## Evidence

```
cargo test -p cmtrace-open --lib single_instance            -> 6 passed
cargo clippy -p cmtrace-open --all-targets -- -D warnings   -> clean
npx vitest run src/hooks/use-file-association.test.tsx      -> 20 passed
npx tsc --noEmit                                            -> clean
```

RED before the module body existed: `cargo test -p cmtrace-open --lib single_instance` failed to compile with `cannot find function parse_forwarded_open_request` / `is_replacement_launch` at eight call sites.

## Deliberately not done

- No verification on Windows. File-association double-click, focus/raise behaviour and the Constrained-Language-Mode-adjacent elevation path can only be confirmed on a Windows machine with the MSI installed; that run is outstanding, and this PR does not claim it.
- No change to how a file is parsed or displayed once open, and no other launch path was touched.

## Summary by CodeRabbit

- **New Features**
  - Added single-instance behavior for desktop launches.
  - Opening files or folders while the app is running now activates the existing window and opens the paths in the active workspace.
  - Multiple paths are processed sequentially, preserving their order and preventing concurrent startup operations.

- **Bug Fixes**
  - Elevated restarts continue to replace the existing process without forwarding paths to the original window.
  - Malformed launch data is safely ignored.

- **Documentation**
  - Updated changelog and elevation acceptance guidance for the new launch behavior.


adamgell:
@coderabbitai review

Both major findings accepted and fixed; the test-only finding is fixed too. The em dash in CHANGELOG.md is retained as house style.

**Working directory.** Resolved in Rust at parse time, as you suggested: the plugin callback passes the second launch's directory into `parse_forwarded_open_request`, and `resolve_forwarded_path` joins a relative path onto it while returning an absolute path byte-for-byte. When the launch reports no directory the path is left as it arrived. The payload keeps its `{ paths: [...] }` shape, so the frontend still receives absolute paths only.

Red: `a forwarded relative path must be resolved before it is opened, got "Logs/ccmexec.log"` → green.

**Serialization.** Forwarded opens now run through one promise queue, with a recovery step so an unexpected failure cannot stall later launches.

Red: `expected 2 to be 1` — two back-to-back launches both entered the open, peak in-flight 2 → green.

**Non-null assertions.** `secondLaunchHandler!` is gone; `deliverSecondLaunch()` throws a clear failure when nothing is registered for the event.

Verified after the fixes: `cargo test -p cmtrace-open --lib single_instance` 8 passed (the two new cases included), `cargo clippy -p cmtrace-open --all-targets -- -D warnings` clean, `npx vitest run src/hooks/use-file-association.test.tsx` 21 passed, `npx tsc --noEmit` clean, the five-file drag-drop/app-action blast radius 48 passed, and `cargo check --no-default-features` clean for the Lite edition.

Still Windows-only and not claimed here: real double-click routing, foreground focus, and the absence of a second window at runtime.

adamgell:
@coderabbitai review

Finding accepted and fixed, and applied to one more site that had the same shape.

The queue is now shared by every launch open rather than only by forwarded ones: the startup `paths` branch enqueues through `enqueueLaunchOpens`, so a startup open still in flight blocks a forwarded launch instead of racing it, and the elevated-restore branch is enqueued too, since `restoreElevatedSource` opens a source exactly like the other two.

The ref is `launchOpens`; `enqueueLaunchOpens(open)` returns the queued promise while keeping the chain's own rejection swallowed, so a failed open cannot stall the launches behind it, and the caller still reports its own failure — the existing `[startup] failed to handle launch intent` log continues to cover a failed startup open.

Red first: the new test failed with the forwarded open started while the startup open was still pending, asserting peak in-flight 1 and order `["C:\\Logs\\startup.log", "C:\\Logs\\forwarded.log"]`.

Green: `npx vitest run src/hooks/use-file-association.test.tsx` 22 passed, `cargo test -p cmtrace-open --lib single_instance` 8 passed, `cargo clippy -p cmtrace-open --all-targets -- -D warnings` clean, `npx tsc --noEmit` clean, blast radius (drag-drop, app-actions, evidence-bundle dialog) 49 passed.

Still Windows-only and unclaimed: real double-click routing, foreground focus, and the runtime absence of a second window.

adamgell:
@coderabbitai review

Finding accepted and fixed, using the precedent the first-launch path already set rather than a new channel.

The paths are now recorded in `AppState` (`second_launch_paths`) before anything is announced, the event carries no payload and only wakes the frontend, and a new `take_second_launch_paths` command takes and clears the list — the same take-and-clear shape as `get_initial_file_paths`. Exactly-once is structural rather than a dedup step: a claim empties the list. The hook registers its listener first and then claims once registration resolves, which is what closes the startup gap, and every claim runs through the existing launch queue so a path waits in the backend until the window can open it.

Consequence worth noting: `ForwardedOpenRequest` and its JSON-shape assertions are gone, because nothing serialises that struct any more. The wire contract is the claim's string array, decoded by the existing `decodeStringArrayResponse`.

Red first: the rewritten frontend tests failed 6 of 23 against the old payload delivery, and the Rust side did not compile against the missing state field and command.

Green: `cargo test -p cmtrace-open --lib single_instance` 10 passed (including a mock app with no main window, asserting the paths are waiting and a second claim finds nothing), `npx vitest run src/hooks/use-file-association.test.tsx` 23 passed, `npx vitest run src/lib/commands.test.ts` 42 passed, `npx tsc --noEmit` clean, `cargo clippy -p cmtrace-open --all-targets -- -D warnings` clean, `cargo check` with and without default features.

Still Windows-only and unclaimed: real double-click routing, foreground focus, and the runtime absence of a second window.

adamgell:
CI caught one more thing after the startup handoff landed, and it is fixed in a follow-up commit.

`E2E (Playwright)` failed on the smoke test, which asserts the app loads without console errors. The new claim command had no entry in the browser shim (`e2e/fixtures/tauri-shim.ts`), so it fell through to the shim's unknown-command default of `null`, the decoder rejected it, and the app logged:

```
[file-association] failed to claim a forwarded launch
{error: Command 'take_second_launch_paths' returned an invalid response}
```

The shim is the model of the backend for these tests, so it now answers `take_second_launch_paths` the way it already answers `get_initial_file_paths` — with an empty list, since a browser fixture never has a second launch. Reproduced locally before the fix and green after: `npm run test:e2e` 21 passed, `npx tsc --noEmit` clean.

adamgell:
@coderabbitai review

Finding accepted and fixed.

The startup effect now enqueues one task holding the three IPC reads and the branch that follows them (file open, elevated restore, or workspace view), synchronously when the effect runs — nothing is awaited before the enqueue, so the slot is taken before the forwarded effect can queue its claim. Startup-before-forwarded ordering is now structural rather than timing-dependent, and the mount claim simply waits behind the startup task. Recovery is unchanged, the existing `[startup] failed to handle launch intent` log still covers a failed startup open, and the elevated-restore branch stays inside the same task instead of nesting another enqueue.

Red first: `reserves the startup slot before a forwarded launch can take it` failed with both claims running ahead of the startup task while the reads were pending.

Green: `npx vitest run src/hooks/use-file-association.test.tsx` 24 passed, `npm run test:e2e` 21 passed (the shim answers the claim command now), `cargo test -p cmtrace-open --lib single_instance` 10 passed, `cargo clippy -p cmtrace-open --all-targets -- -D warnings` clean, `npx tsc --noEmit` clean, six-file blast radius 93 passed.

Two notes for the record: two older tests needed updating because the forwarded claim now waits behind the whole startup task rather than only its open; and this file's own mocks had a leak — per-test `mockResolvedValueOnce` queues survived `vi.clearAllMocks()`, which is why the claim-failure test passed in isolation and failed in a full run. `beforeEach` resets that mock now.

adamgell:
@coderabbitai review

No review has run against the current head (`2c2580a1`) since the last push, and the recorded decision is from an earlier head. All findings from the previous cycle are addressed or dispositioned with zero unresolved threads; a fresh pass at this head is what the gate needs.

adamgell:
@coderabbitai review

adamgell:
## Verified against the real app, not the shim

The one claim no automated gate covers is the actual hand-off: the e2e suite runs against a browser shim, so nothing here had exercised the real single-instance plugin. Done now, on Windows, with a debug build of this head (`tauri-plugin-single-instance v2.4.4`, frontend served by Vite on :1420, WebView2 CDP on :9222 so the window's own DOM could be read).

**Before** — the window on the Log Explorer workspace, no source:

```
title: "CMTrace Open"   tabs: 0   hasNeedle: false
"... Workspace: Log Explorer ... NO SOURCE  Open a log file or folder ..."
```

**Second launch** — `cmtrace-open.exe C:\Users\…\Documents\verify-565-forwarded.log`, i.e. what a file-association double-click produces:

```
second launch: exited=True  code=0  after 178 ms
processes after: 1
```

**After** — the same running window, with no second window anywhere:

```
tabs: 1   hasNeedle: true
"... verify-565-forwarded.log   FILE SOURCE   verify-565-forwarded.log
    C:\Users\…\Documents\verify-565-forwarded.log"
```

So on this machine: the second process routed its argument and exited in 178 ms, exactly one process remained, and the window that was already open raised itself and opened the path in a tab. That is #565's acceptance, observed rather than inferred.

Gates on this head for the record: all 15 checks pass, including the three platform builds, Check & Test, E2E and ESP Diagnostics; `cargo check --all-targets`, `npx tsc --noEmit` and `npm run test` (1567 tests) clean locally on the resolved merge.

The elevated-replacement path (a `--restore` relaunch must *not* route itself) is not covered by the above — it needs an elevated launch — and remains pinned by unit tests in `single_instance.rs`.

PR #624, opened 2026-09-15
chore(scoop): bump bucket to 1.6.0

adamgell:
Points the checked-in Scoop manifests in `bucket/` at v1.6.0.

Produced by `.github/workflows/scoop-publish.yml` when the release event for v1.6.0 fired: the job ran `.github/scripts/Update-ScoopManifest.ps1` against the release's own autoupdate templates, then installed and uninstalled both manifests with real Scoop on a Windows runner, so the hashes, the download URLs and the shim wiring are validated before this lands.

The pull request is opened by hand because the workflow cannot open one yet: `GITHUB_TOKEN` may only create pull requests when the repository allows it, and this repository currently has "Allow GitHub Actions to create and approve pull requests" disabled. The job pushed the branch and failed at `gh pr create` with `GitHub Actions is not permitted to create or approve pull requests`. PR #623 changes that job to open the pull request itself once the automation is permitted; until then this is the manual hand-off.

Merging this unblocks `scoop install cmtrace` / `scoop install cmtrace-lite`, which have been serving the last hand-bumped version.


adamgell:
Superseded: the workflow can open this pull request itself now that the repository allows GitHub Actions to create pull requests. Re-dispatching the publisher to prove the create path end to end; it will push the same branch and open the replacement.

PR #638, opened 2026-09-19
fix(build): restore green main after the dependency batch

adamgell:
## What this is

`main` is red on every axis since the dependency batch merged today (CI runs #1721–#1729 all failed). This restores green main with the smallest possible changes: pin the three bumped dependencies back to the versions the code was written against. The migrations are real work and need deliberate adaptation, not dependabot bumps.

## The three breakages

| Dep | Bump | Breakage |
|---|---|---|
| `quick-xml` | 0.41 → 0.42 | Changed event API (`BytesText::into_inner` returns `Cow<str>`, `GeneralRef::decode` gone, attribute decoding deprecated) → breaks `event_node.rs`, `writer.rs`, `intune/evtx_parser.rs` |
| `windows-future` | 0.3.2 → 0.100.0 | Built against windows-core 0.100 while the app pins `windows 0.62.2` → duplicate windows-core → `IAsyncOperation` trait-bound mismatch in `graph_api.rs` |
| `react` / `react-dom` | 19.2.8 → 19.3.0 | Changes module resolution so vitest sees `@fluentui/react-tabster` as CommonJS → every suite importing a theme fails (`createTabster` / `useFocusVisible` named exports) |

Bisected: pinning react back to 19.2.8 alone fixes vitest with all other frontend bumps kept. Pin `@fluentui/react-tabster` was tried first and does **not** fix it.

## Changes

- `src-tauri/Cargo.toml`: `quick-xml = "=0.41.0"`, `windows-future = "=0.3.2"` (exact pins with comments recording why).
- `package.json`: `react` / `react-dom` pinned to `19.2.8`.
- Regenerated `Cargo.lock` and `package-lock.json` accordingly.

## Verified locally

- `cargo check --locked --all-targets` ✅
- `cargo check --locked --no-default-features` ✅
- `cargo test --locked --jobs 8` full suite ✅
- `cargo clippy --locked --all-targets --jobs 8 -- -D warnings` ✅ (full + lite)
- `npx tsc --noEmit` ✅
- `npm run test` ✅ 133 files / 1553 tests
- CI script node tests ✅ (bundle outputs, release checkout, toolchain)

## Caveats

- `scripts/ci-windows-provenance.test.mjs` fails locally on this machine (8.3 short-name path comparison, `ADAM_A~1.HOM` vs `adam_admin.HOME`). It fails identically on the pre-batch tree and is unrelated to these pins; CI runs the node tests on Ubuntu.
- Dependabot will re-propose these bumps; they need a deliberate migration (quick-xml 0.42 API, windows 0.100 family alignment, react 19.3 module-resolution check), tracked separately.

## Summary by CodeRabbit

* **Chores**
  * Updated application and platform components to specific compatible versions.
  * Added compatibility notes for selected platform integrations.


adamgell:
Superseded by #641, which restores the same three dependency pins (quick-xml =0.41.0, windows-future =0.3.2, and the pre-bump npm production dependencies) directly on main. Closing to avoid duplicating that change; the other Event Viewer PRs will rebase onto main as fixed by #641.

